PR #7, opened 2020-06-11
support search_disabled config

PeculiarGoat:
Happy with this?


site/templates/nav.jet
```diff
@@ -1,11 +1,13 @@
-{{block nav(showSubNav=true)}}
+{{block nav(showSubNav=true, searchDisabled=config("search_disabled"))}}
```
additiverich, line 1: 1/ Does `searchDisabled` need to be an option here? Can we rely on checking the config inside the block?

2/ If it needs to be here, it relies on someone to have set up the config in uber, so maybe have a default value of "false" passed into `config`

PeculiarGoat: Yeah I don't know why I did that.
It's better now.

PR #10, opened 2020-06-14
Powered by

gscragg:
Stop text wrap when footer links are larger than expected.
Split into separate template so easier to change.

Note usage of `{{include "footer/powered-by" }}` instead of previous combo of `{{ import... }}` and `{{yield ...() }}`.

Seems a bit easier to manage.

### DIFF
--- a/site/styles/_footer.scss
+++ b/site/styles/_footer.scss
@@ -80,6 +80,7 @@ footer {
 		margin-left: auto;
 	}
 	a {
+		white-space: nowrap;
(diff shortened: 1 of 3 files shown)


site/styles/_footer.scss
```diff
@@ -80,6 +80,7 @@ footer {
 		margin-left: auto;
 	}
 	a {
+		white-space: nowrap;
```
stajs, line 83: Nice, I've been manually doing this when I remember.

PR #11, opened 2020-06-15
Style

cunhapaola:


site/styles/_pages.scss
```diff
@@ -15,6 +15,9 @@
 .page-header {
   padding-right: 20px;
   padding-left: 20px;
+  @media (min-width: 438px) {
```
additiverich, line 18: 438px is a bit of an arbitrary breakpoint which doesn't match any of the widths specified in the variables file and I wonder what problem this tries to solve.

site/styles/_nav.scss
```diff
@@ -288,12 +288,16 @@
...
   }
 }
+
 // Header Form Search
 .navbar-nav-search {
   position: absolute;
   left: 160px;
   width: 150px;
   top: $navbar-search-top;
+  @media (max-width: 345px) {
```
additiverich, line 298: See my comment elsewhere about arbitrary break points (although this is quite close to the xxs width which is 320px so maybe this could be used here?)

site/styles/_footer.scss
```diff
@@ -40,7 +42,7 @@ footer {
 		}
 	}
 	.copyright-statement {
-		font-size: 8px;
+		font-size: 12px;
```
additiverich, line 45: 👍  Always find myself increasing this size because I'm old and can't read tiny print

stajs:
What is this PR trying to achieve? I can see changes but have no context for them. Can you please add a description of what problem you are trying to solve. I assume you are working off a JIRA card so it should also have the JIRA number in the title.

PR #22, opened 2020-07-12
Update carousel.jet

additiverich:
Make it easier to remove pagination controls from the carousel when removing individual slides by querying them by slug


additiverich:
> Is this for the case where the carousel has a single item? Does it not already hide the controls?

It's for when titles you're aren't allowed to watch are dynamically removed from the page (something 2 sites do now), including the carousel.

PR #28, opened 2020-08-12
fixed hamburger bug

PeculiarGoat:
No more error in console.  X icon works like the one on Lonely Planet.

### DIFF
--- a/site/static/js/main.js
+++ b/site/static/js/main.js
@@ -421,17 +421,9 @@ function initializeMenu() {
 }
 
 function toggleMobileMenu(icon){
-  var element = document.getElementById('navbar-mobile');
-  if(element){
-    if(element.classList.contains('visible')) {
-      if(icon) icon.classList.remove('toggled');
-      element.classList.remove('visible');
-    }
-    else
-    {
-      element.classList.add('visible');
-      if(icon) icon.classList.add('toggled');
-    }
+  var element = document.querySelector('.collapse.navbar-collapse');
+  if (element && icon) {
+    element.classList.contains('show') ? icon.classList.add('toggled') : icon.classList.remove('toggled');
   }
 }
 
--- a/site/templates/nav.jet
+++ b/site/templates/nav.jet
@@ -13,7 +13,7 @@
         </form>
       {{end}}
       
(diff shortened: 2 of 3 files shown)


site/templates/nav.jet
```diff
@@ -13,7 +13,7 @@
         </form>
       {{end}}
       
-      <button class="navbar-toggler" type="button" s72-collapse=".navbar-collapse" aria-controls="navbarsExampleDefault" aria-expanded="false" aria-label="Toggle navigation">
+      <button class="navbar-toggler" type="button" toggle-function="toggleMobileMenu" s72-collapse=".navbar-collapse" aria-controls="navbarsExampleDefault" aria-expanded="false" aria-label="Toggle navigation">
```
gscragg, line 16: Where does this new attribute get used?

PeculiarGoat: [Here](https://github.com/shift72/core-template/blob/51d1dd5ba0ad9d508d944b75608d96cc1d3c4451/site/static/js/main.js#L407).

site/static/js/main.js
```diff
@@ -421,17 +421,9 @@ function initializeMenu() {
...
-      if(icon) icon.classList.remove('toggled');
-      element.classList.remove('visible');
-    }
-    else
-    {
-      element.classList.add('visible');
-      if(icon) icon.classList.add('toggled');
-    }
+  var element = document.querySelector('.collapse.navbar-collapse');
+  if (element && icon) {
```
stajs, line 425: This is nicer from the "less lines of code" POV, but is it possible to not have an icon? The previous more verbose code did cater for that situation.

PeculiarGoat: It didn't... it checked for icon in the if and the else.

stajs: Yeah, my poor phrasing - let me rephrase. Before, it still added or removed a class on the element (and didn't matter if there was an icon), is that adding/removing class on the element no longer required (or handled elsewhere)?

PeculiarGoat: Ah right... yeah it's no longer required.  This js file was copy/pasted (I assume) from elsewhere but the template containing the elements that it selects wasn't so the `show` class is used instead of the `visible` class and seems to be handled elsewhere.

PR #29, opened 2020-08-13
AB#561 more IT language variables

PeculiarGoat:
Just missing the playback window message for the player... I can't figure out where that one is hidden (not in core-template/relish/the player)!


stajs:
That message might be one where the key is concatenated so you can't search for the whole key to find it.

PR #32, opened 2020-08-18
More language vars

PeculiarGoat:
I've added all of the i18n strings from Relish that I can find and put TODO in places where I don't know what the value should be.  I'm guessing those should all be filled out before this gets merged... but they had no default values in Relish so maybe it's not a big deal.

### DIFF
--- a/site/en_AU.all.json
+++ b/site/en_AU.all.json
@@ -366,10 +366,10 @@
 
   "plan_label_owned": { "other": "" },
 
-  "pricing_quality_hd": { "other": "TODO" },
-  "pricing_quality_sd": { "other": "TODO" },
-  "pricing_quality_hd_only": { "other": "TODO" },
-  "pricing_quality_sd_only": { "other": "TODO" },
+  "pricing_quality_hd": { "other": "pricing_quality_hd" },
+  "pricing_quality_sd": { "other": "pricing_quality_sd" },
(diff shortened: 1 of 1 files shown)


site/en_AU.all.json
```diff
@@ -352,5 +356,81 @@
...
+  "usersubscriptions_subscription_status_past_due": { "other": "TODO" },
+  "usersubscriptions_subscription_status_trialing": { "other": "TODO" },
+  "usersubscriptions_subscription_unsubscribe": { "other": "TODO" },
+  "usersubscriptions_unsubscribe_modal_cancel": { "other": "TODO" },
+  "usersubscriptions_unsubscribe_modal_confirm": { "other": "TODO" },
+  "usersubscriptions_unsubscribe_modal_title": { "other": "TODO" },
+  "usersubscriptions_unsubscribe_modal_body": { "other": "TODO" },
+
+  "shopping_card_update_reason_none": { "other": "TODO" },
+  "shopping_card_update_reason_ + this.updateCardReason": { "other": "TODO" },
```
stajs, line 416: Looks like the various `updateCardReason`'s need to be enumerated

PeculiarGoat: I'm going to remove that one... updateCardReason comes from a query string and I can't find an example of it being used anywhere.

stajs: Sweet

stajs (approved):
:shipit:

PR #37, opened 2020-08-28
Sprichst du Deutsch

PeculiarGoat:
Just copied loads of German from JFF to core-template so other sites like BIFF can use it.


stajs (approved):
I don't think this is worth a PR in the future, just bang 'em in

PR #40, opened 2020-09-02
Show suspended message instead of redirecting

gscragg:
Also show a nicer message if IP throttled after too many failed attempts.

In preparation for a relish release (1.3.33).


stajs (approved):
Sorry I missed this

PR #41, opened 2020-09-04
built in festival scope footer

PeculiarGoat:
Set user config `festival_scope_footer` to `true` and public to turn on Festival Scope footer for a site.


stajs (approved):

site/styles/_footer.scss
```diff
@@ -94,4 +94,27 @@ footer {
...
+// Festival Scope
+.footer-festivalscope {
+  font-size: .8rem;
+
+  a {
+    padding: 3px 0 6px 0;
+    margin: 0;
+
+    &:hover .footer-festivalscope-logo {
+      filter: invert(100%);
```
stajs, line 107: Not sure about this though

PeculiarGoat: Not sure in what way?

stajs: Ignore... wasn't sure how it would look in practice and in particular how the inversion looked on different client colours, but happy enough that will have to be tweaked for each client rather than a silver bullet here.

PR #42, opened 2020-09-09
Greek/translation

smitsanghvi21:
Added the Greek translation based on [this](https://github.com/indiereign/shift72-thessaloniki/blob/master/local/site/en_AU.all.json) file.


PeculiarGoat (approved):
Nice work @smitsanghvi21

PR #45, opened 2020-09-11
bonus content fix

smitsanghvi21:
- Removed `d-flex` class from the main div.

- Replaced `slider` with `list`.

- Added css from the `faclon`


stajs (approved):
Much easier to understand, thanks!

PR #47, opened 2020-09-22
fix long string wrapping issue

smitsanghvi21:
ADO: [Click here](https://dev.azure.com/S72/SHIFT72/_workitems/edit/838/)

image

### DIFF
--- a/site/styles/main.scss
+++ b/site/styles/main.scss
@@ -345,3 +345,10 @@ s72-classification-label {
 		width: 100%; // otherwise inherits from meta_item post which has a fixed width
 	}
 }
+
+//text wrap fix
+@for $index from 1 through 6 {
+  h#{$index}, p {
+      overflow-wrap: break-word;
+  }
+}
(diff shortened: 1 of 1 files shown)


stajs (changes requested):
I don't think it's worth the loop and it's going to repeat the `p` each time too.

So instead of generating:

```css
h1, p {
}

h2, p {
}

... etc ...
```

... I think better to just be plain with:

```css
h1, h2, h3, h4, h5, h6, p {
}
```

site/styles/main.scss
```diff
@@ -345,3 +345,8 @@ s72-classification-label {
 		width: 100%; // otherwise inherits from meta_item post which has a fixed width
 	}
 }
+
+//text wrap fix
+h1, h2, h3, h4, h5, h6, p{
```
stajs, line 350: Maybe just a space after between the `p` and `{`.

stajs (approved):

PR #48, opened 2020-09-23
increases page padding when subnav wraps

PeculiarGoat:
Bumps the whole page down by nav-item height if the subnav wraps.  Bumps it back up if the subnav unwraps.

### DIFF
--- a/site/static/js/main.js
+++ b/site/static/js/main.js
@@ -492,6 +492,7 @@ function documentReady(app) {
   });
 
   togglePageTopPadding();
+  window.addEventListener('resize', togglePageTopPadding, { passive: true });
 }
 
 function detectTouchscreen(){
@@ -507,5 +508,3 @@ document.addEventListener('s72loaded', function(event) {
   let app = event.detail.app;
   documentReady(app);
 });
-
-window.onresize = togglePageTopPadding;
--- a/site/static/js/main.js
+++ b/site/static/js/main.js
@@ -460,6 +460,19 @@ function togglePageTopPadding() {
   }
 }
 
+function throttle(callback, limit) {
+  var waiting = false;
+  return function() {
+    if (!waiting) {
+      callback.apply(this, arguments);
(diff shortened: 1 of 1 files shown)


site/static/js/main.js
```diff
@@ -472,3 +507,5 @@ document.addEventListener('s72loaded', function(event) {
   let app = event.detail.app;
   documentReady(app);
 });
+
+window.onresize = togglePageTopPadding;
```
gscragg, line 511: Please use `window.addEventListener` and move it into `documentReady()`

gscragg: You should also do your best to ensure this function isn't called every cpu cycle as the user resizes.
Without any type of throttling this function will run 100s of times per resize.

The slider and - maybe - carousel have examples on throttling resize logic.

site/static/js/main.js
```diff
@@ -427,6 +427,39 @@ function toggleMobileMenu(icon){
...
+
+function togglePageTopPadding() {
+  var navItem = document.querySelector('.sub-nav .nav-item');
+  if (!navItem) return;
+
+  var page = document.querySelector('.page');
+  if (!page) return;
+
+  if (subnavOverflowing()) {
+    if (page.hasAttribute('style')) return;
```
gscragg, line 455: This implies that we are only supporting a single resize?

gscragg (changes requested):

site/static/js/main.js
PeculiarGoat: This prevents the padding from growing 100s of times per resize.  I realise that this check being indirect isn't a great way.  Maybe I could hold onto the page padding value in a global variable instead?

gscragg (changes requested):
See if the `s72.utils.fn.throttle()` suits your needs and if so use that instead of rolling your own.

gscragg (approved):

PR #51, opened 2020-09-28
fix expired translation bug

smitsanghvi21:
ADO card: https://dev.azure.com/S72/SHIFT72/_workitems/edit/1332/

The site is https://online.hardline-festival.de/ and not Hamburg. Looks like {{.Expiry}} tag was showing up in the "Expired' translation. If I am not wrong, the date shouldn't be showing up and hence removed the tag. Verified it using the English translation for staging-abc and just the "expired" tag was showing up without any date.

### DIFF
--- a/site/de_DE.all.json
+++ b/site/de_DE.all.json
@@ -356,7 +356,7 @@
 
   "availability_coming_soon": {"other": "In Kürze verfügbar"},
   "availability_available": {"other": "Verfügbar ab {{.ReleaseDate}}"},
-  "availability_expired": { "other": "Abgelaufen {{.Expiry}}" },
+  "availability_expired": { "other": "Abgelaufen" },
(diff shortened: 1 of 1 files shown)


smitsanghvi21:
@stajs Once you get time can you please review this? :)

stajs (approved):
This is the best we can do for now.

Just to be aware there is an underlying issue with shift72-web where the availability label and status are sharing the same translation key but only one of them is piping through the date. I view this as a bug - if we ever override the default value this will happen until that bug is fixed.

image

timhowellvista:
Another workaround for this. For English sites, you can remove the availability_expired line completely from the language json file and it falls back to default with the availability status not having the date and the availability label having the date.

Though I still consider this bug unfixed until you can use {.Expiry}} in a non-English site

PR #52, opened 2020-10-05
Sticky footer

curtisadams28:


site/styles/main.scss
```diff
@@ -52,6 +52,9 @@ html {
 }
 body {
 	font-family: "Rawline", $font-family-base;
+	min-height: 100vh;
```
stajs, line 55: How much testing has this had across browsers and breakpoints? I'm cautious as I remember being burned by these units in the early days of browser support. @brynyShift72

site/styles/_pages.scss
```diff
@@ -70,8 +72,6 @@
   }
 }
 .form-page {
-  margin: 0 auto;
-  max-width: 1600px;
```
stajs, line 74: We don't want this constraint anymore?

site/styles/_pages.scss
```diff
@@ -3,6 +3,8 @@
   @include media-breakpoint-up(md){
     padding-bottom: $grid-gutter-width;
   }
+  flex: 1 0 auto;
```
stajs, line 6: Is this doing the same as the removed `margin: 0 auto`?

site/styles/_footer.scss
```diff
@@ -2,6 +2,7 @@
 footer {
 	padding: 10px 10px;
 	flex-direction: column;
+	flex-shrink: 0;
```
stajs, line 5: I'm not familiar with this one, explanation?

stajs (changes requested):
Is there an ADO card for this work so @brynyShift72 can test it?

Can you do some before and after screenshots so we can visualise.

site/styles/_pages.scss
curtisadams28: It is shorthand for: flex-grow, flex-shrink, flex-basis. The .page class is only as high as its content. Setting the .page to ‘flex-grow: 1’ makes it grow to take up all of the remaining space in the parent flex container (body). This is what pushes the footer to the bottom of the page.
(https://cssreference.io/property/flex-grow/)

site/styles/_pages.scss
curtisadams28: The margin: 0 auto was was squashing the width of the forms once I changed the page to flexbox. Once the max-width reached 1600px, the .form-page container would not extend across the width of the page. This resulted in the form being off centre. 

image

site/styles/_pages.scss
curtisadams28: I was using the flexbox section of this article here for the fix (https://css-tricks.com/couple-takes-sticky-footer/). From my testing however, it seems like flex-shrink is unnecessary (as well as the auto).

gscragg:
Shelter has an example of a sticky footer, although not truly sticky; it wont go higher than the bottom of the screen if the content is too short.

site/styles/_pages.scss
stajs: Cool

site/styles/_pages.scss
stajs: Okay, my comment was that `margin: 0 auto` is used to center the content - if it's removed, doing it some other way?

curtisadams28: It is still centered because there is already another `margin: 0 auto` inside the form container. I have only removed it from the parent. I did this because when I changed `.body` to flexbox, `margin: 0 auto` was squashing the form.

site/styles/_footer.scss
curtisadams28: flex shrink is the opposite of flex grow. When the content is too large for a container, it tells sibling flex items the rate at which they should shrink. Eg. If item1 has a shrink value of 1 and item2 has a shrink value of 2, item2 will shrink twice as much as item1. By default, the value is 1. Changing it to 0 means that flex items will not shrink smaller than their default size. I think what this does is prevent unwanted vertical scaling (remember that .page and footer are now flex items in a column).

site/styles/_pages.scss
stajs: Gotcha 👍

site/styles/_footer.scss
stajs: Thanks for the ELI5! 👍

stajs:
Fixes AB#1436

stajs (approved):

smitsanghvi21 (commented):
Works fine when connected to desktop, but on default zoom (100%) on MacBook Pro 15 inch when not connected to monitor it doesn't. It works fine when the zooms is less than 80%. 

image

PeculiarGoat:
I'm a little bit sceptical of the need to mess around with flex.

I did this on Neon + Indy with just:
```
.page {
  @include media-breakpoint-up(md) {
    min-height: calc(100vh - 89px);
  }
}
```
Where 89px is the height of the footer (that would need to be obtained dynamically if this were used in core-template).

stajs:
@PeculiarGoat Fair point, but I guess the trade-off is how to do the dynamic obtainment? We could have another var for footer height but becomes another point of friction.

PeculiarGoat:
@stajs I guess it can't be dynamic without javascript so that's probably not a great idea.  But adding a var for footer height is probably not such a big deal (it's the same as using nav height for page top-padding).

curtisadams28:
I accidently saved the conflicts together but I've fixed it in the last commit.

site/styles/_pages.scss
```diff
@@ -69,9 +71,10 @@
     min-height: 740px;
   }
 }
+.account-page {
```
PeculiarGoat, line 74: I'm curious why this needs to exist... if the signin/signup pages can exist with just the `form-page` class, why can't the account page?

curtisadams28: Yup. You're right. It doesn't need to exist. I think I previously had `form-page` as `margin: 0 auto` set and it was causing the account page to bug out. Removed in the latest commit.

site/styles/_pages.scss
stajs, line 6: I believe [this](https://github.com/philipwalton/solved-by-flexbox/blob/master/assets/css/components/site.css#L21) should be just `flex: 1;` (as per [example](https://philipwalton.github.io/solved-by-flexbox/demos/sticky-footer/)) with `0 auto` redundant because of the aforementioned browser fixes.

site/styles/main.scss
```diff
@@ -52,7 +52,10 @@ html {
   overflow-y: scroll;
 }
 body {
-  font-family: "Rawline", $font-family-base;
+	font-family: "Rawline", $font-family-base;
+	min-height: 100vh;
```
stajs, line 56: [Apparently there is an IE11 `min-height` bug](https://github.com/philipwalton/solved-by-flexbox/blob/master/assets/css/components/site.css#L2), so [potentially should just be `height`](https://github.com/philipwalton/solved-by-flexbox/blob/master/assets/css/components/site.css#L12) - has this been tested there?

site/styles/_meta-detail.scss
```diff
@@ -66,6 +66,7 @@
     right: 0px;
     margin-left: auto;
     margin-right: auto;
+    height: 100%;
```
stajs, line 69: Why do you need this with the `body` being `100%` `flex`?

stajs (commented):
I know this has been talked about a lot, but if we are using that canonical example I just want to understand these differences.

site/styles/main.scss
curtisadams28: Yeah. The `min-height:` does not work on IE11. By using `height: 100%`, you also need to add `height: 100%` to the html element. This is because the body relies on a parent for the height. In addition, replacing `min-height` with `height` also means that the body doesn't extend down the page anymore.

image

Is this important to consider? or is it a non-issue?

site/styles/_footer.scss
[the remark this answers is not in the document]
curtisadams28: I will change in next commit.

site/styles/_meta-detail.scss
curtisadams28: This was due to an issue where the height of the background was set larger than the height of the body at some resolutions. 
 
image

I did `height: 100%` so that it is only ever as tall as its parent (.page) until it reaches `max-height: 900px`. This works in IE11.

site/styles/_pages.scss
curtisadams28: will change in next commit

site/styles/main.scss
stajs: I guess the important factor is, if `min-height` doesn't work on IE11, does that mean the sticky footer doesn't work on IE11?

curtisadams28: The sticky footer does not work in IE11 with `min-height` but it does work with `height`. I am going to use the latter method.

PR #54, opened 2020-10-13
fixed gradient issue for detail background images

PeculiarGoat:
nothing to it


stajs:
Sure is simple change, but what was the problem?

stajs:
Ah, I couldn't see a link to an issue but then spotted your branch name it was hiding in there

stajs (approved):

PR #57, opened 2020-10-15
AB#1620 moved wishlist script into main.js

PeculiarGoat:
This bug is an unexpected side-effect of adding rollup.js.
The script in the wishlist file was trying to call the `initializeSwiper` function which now get encapsulated by rollup so was causing an error.  I've fixed it by moving the wishlist code into main.js (and checked that there are no similar problems elsewhere).

I think all shift72-template sites that have been created since I did the ES6 stuff will need core updating after this is merged.

### DIFF
--- a/site/static/js/main.js
+++ b/site/static/js/main.js
@@ -1,5 +1,37 @@
 var slideObservers = [];
 
+function initializeWishlist() {
+  var wishlist = document.querySelector('s72-userwishlist');
+  var originalFunction = wishlist.classList.remove;
+  wishlist.classList.remove = function(className){
+    if(className == 's72-hide')
+    {
+      // Hide this from view
+      wishlist.style.opacity = '0';
+
+      // Remove class
+      originalFunction.apply(this, [ className ]);
+
+      /* Convert this to a swiper */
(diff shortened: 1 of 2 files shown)


site/static/js/main.js
```diff
@@ -1,5 +1,37 @@
 var slideObservers = [];
 
+function initializeWishlist() {
+  var wishlist = document.querySelector('s72-userwishlist');
+  var originalFunction = wishlist.classList.remove;
```
stajs, line 5: Do we need a null check / guard?

site/static/js/main.js
```diff
@@ -1,5 +1,37 @@
...
+        var swiper = initializeSwiper(container, true);
+
+        init(swiper);
+        toggleButtons(container);
+      }
+
+      // Now show it
+      wishlist.style.opacity = '1';
+    }
+    else
```
stajs, line 28: Why the else? You always going to call the original func?

stajs (approved):
Approving because of urgency, but I dislike overriding the remove function - perhaps there is a better way when the fire is out.

PR #60, opened 2020-11-03
Add lang code to html links

PeculiarGoat:
I've only done the German language file so far as a proof of concept... no point doing the rest until everyone is happy with the method.

- Updated kibble to `0.15.0` so we can use the new `routeToPath` helper function (it adds the correct language prefix to any path passed into it).
- Replaced hard-coded hrefs in jet templates with `routeToPath` calls e.g. `<a href="{{ routeToPath("/signin.html") }}">`.
- Replaced hard-coded Relish component attributes with `routeToPath` calls e.g. `<s72-acceptinvite-form acceptinvite-url="{{ routeToPath("/acceptinvite.html") }}" />`.
- Replaced hard-coded hrefs in German language file with variables that will be passed in from Relish e.g. `"userlibrary_empty_content": { "other": "Durchsuchen... <a href=\"{{ .HomeURL }}\">Filmauswahl</a>, um Ihre..." }`.
- Added Relish component attributes to feed in paths that will be used in language file variables e.g. `<s72-cookie-consent policy-url="{{ routeToPath("/page/cookie-policy/") }}">`.  I chose to do this so we can keep control of the paths in the template.  For example, a German client might want a cookie policy page with a German slug (`/page/cookie-richtlinie/`) which wont work if Relish has baked in default path values (even if they are correctly setting the language prefix).

Thoughts @gscragg @stajs ??

Assuming this is all okay, next steps are:
- Modify all of the Relish components that need to accept new path attributes and feed variables into language strings.
- Modify all of the Relish components that output elements with hard-coded hrefs e.g. `<span dangerouslySetInnerHTML={ { __html: i18n('shopping_card_change', 'Credit card not correct? <a href="/subscriptions.html">Click here to update your card.</a>')} } />` from `change-subscription-payment-method.component.js`.  Should be pretty easy to insert the language prefix everywhere that this happens.
- Go through the other language files in core-template and replace all of the hard-coded hrefs with variables to be passed in from Relish.
- Add the language selector component to Relish.
- Hope that I didn't miss anything.

Given how slowly this has been progressing, I might copy the dirty hacks language selector from Bifed to Los Cabos tomorrow so it's there as a backup if I don't get this all finished and deployed in time.


stajs:
I like where this is going. I think we need to see the other side with a Relish PR, and thinking ahead - how to keep it non-breaking by ensuring components have defaults or ignore null path attributes for older site builds.

site/templates/application.jet
```diff
@@ -41,7 +41,8 @@
     <script>
       window.addEventListener('DOMContentLoaded', function(){
         s72.cfg(function(){ return s72.i18n.load('{{lang.Code}}', '/{{lang.DefinitionFilePath}}'); });
-        s72.boot({ config: {{ site.Config | json | raw }}, toggles: {{ site.Toggles | json | raw }} });
+        s72.cfg(function(app){ return s72.href.load('{{site.SiteConfig.DefaultLanguage}}', '{{lang.Code}}'); });
```
stajs, line 44: Nit: don't pass in `app` if not using?

PeculiarGoat: True!

PeculiarGoat:
I reckon this is done... but can't be merged until https://github.com/indiereign/shift72-web/pull/149 is in prod.

stajs (approved):
Merge after https://github.com/indiereign/shift72-web/pull/149

PeculiarGoat:
Note before I forget:
The current styling/location of the language selector component sucks so get design team to have a look before merging!

kibble.json
```diff
@@ -10,7 +10,7 @@
     "fr": "fr_FR",
     "de": "de_DE",
     "es": "es_ES",
-    "es_MX": "es_MX",
+    "es-mx": "es-mx",
```
stajs, line 13: Not sure how I feel about this being the only value that's hyphenated. Possibly live with it if the aim is to standardise over time.

PeculiarGoat: Yeah I just wasn't sure if I should do them all at once because every site that updates core after this will need the hyphens changing on their local lang filenames.

gscragg: Why can't it be `"es-mx": "es_MX",`?

PeculiarGoat: I'm sure that didn't work when I tried it out before but it does work so I'll change it.

stajs (approved):
Not a fan of the language name duplication, but if it gets us over the line 😬 

Should `es-mx` be `Espanol (Mexican)` or something?

site/templates/application.jet
```diff
@@ -40,8 +40,19 @@
...
+          config: {{ site.Config | json | raw }},
+          toggles: {{ site.Toggles | json | raw }},
+          languages: [
+            {{ range k, v := site.SiteConfig.Languages }}
+              {
+                value: '{{ k != site.SiteConfig.DefaultLanguage ? k : "" }}',
+                name: '{{ i18n(lower("lang_name_" + v)) }}'
+              },
+            {{ end }}
+          ]
```
gscragg, line 54: I am not a fan of this, theres too much reliance on custom mapping here.
We have 2 options:
1. Force relish code to accept the data structure kibble exposes via `site.Languages` or `site.SiteConfig.Languages`.
2. Expose a function in kibble that returns the data you want.

stajs: What if we updated kibble to support this structure in the `kibble.json`?

```
"languages": {  
  "es": { "code": "es_ES", "name": "Español" },
  "es-MX": { "code": "es_MX", "name": "Español (Mexican)" }
}
```

gscragg: My concern with that approach (ignoring some interesting json parsing) is it doesn;t support the case where a Language name is different when referenced by different languages. That could be a non issue but English is not called English in French.

stajs: True, but in the scenario of an English-speaker landing on the French site, they would probably be looking for the "English" option rather than "Anglais". So question is to optimize for the visitor or the language?

gscragg: Sure, lets ride with that as a decision. @PeculiarGoat are you comfortable making kibble changes, or would you like me to look at it?

PeculiarGoat: @gscragg I think option 2 is tidier.
I could stick a function on the site model `func (site *Site) LanguageSelectorOptions() []map[string]string` to return `[{ "value": "en", "name": "English" }, {etc...}]`

@stajs Yeah I think I can do that... could change `models.Config.Languages` to a typed struct and modify the unmarshalling function to handle objects instead of strings.  Then I think it would just be some small changes in the render function.  What do you think @gscragg?

gscragg: > Yeah I think I can do that... could change models.Config.Languages to a typed struct and modify the unmarshalling function to handle objects instead of strings. 

I think if we keep the models as arrays it will be easier for any implementors, the object can have `code` `locale` (or the proper iso or ietf name for it), and `name` properties.

The code that parses kibble.json will need to handle both objects and strings, as when this goes to staging/prod will potentially be used by other templates that haven't updated.

PeculiarGoat: Okay cool I'll have a go!

stajs: Yep good point on back compat.

On the topic of naming things, I propose we try and model on IETF BCP 47 ([the same as browsers](https://www.w3.org/International/articles/language-tags/)) where it would be the `region` subtag. Future advantages could be to not just represent countries, but well, regions.

> The region subtag in RFC 3066 took its values from the ISO 3166 country codes. These two-letter codes are still available from the new registry, but the registry also lists 3-digit UN M.49 region codes. The advantage of these codes is that they can represent more than just countries. For example, localization groups have for some time wanted to label their carefully crafted translations as Latin-American Spanish, rather than the Spanish of any particular country. With RFC 5646 this is possible; the appropriate language tag is `es-419`.

We have had one request for this exact example of Latin-American Spanish so it would be nice to think outside of specific countries and have it available as a core translation.

Other potential future niceness is using the `extlang` and `script` subtags, for places like China where many different dialects are spoken and written.

> For example, `zh` means Chinese, but it covers many Chinese dialects, often mutually incomprehensible. When `zh` is used on its own, it is usually used to mean the predominant language in the encompassed range, although this is not explicitly specified in BCP 47. For example, conventionally `zh` is considered to represent the predominant, Mandarin form of Chinese. Where absolute clarity is needed you can use `cmn` instead as long as that doesn't break interoperability, however, if you are using `zh` to represent a language which is not Mandarin, such as Hakka Chinese, you are better off using the explicit code (in that case, `hak`).

stajs: Got a bit carried away there focusing on the `region` subtag. The whole shemozzle is variously called in the RFC `language tag`, `langtag` or `tag`. I'd be keen to support regions.

PeculiarGoat: @stajs would that mean having files like `es_419.all.json` and `zh_Hans.all.json`?  Do we want to support those right away... it might work with just a bit of regex tweaking in the language selector.

stajs: Lets leave all that extra potential niceness until later and keep this PR as small as possible to solve immediate problem.

gscragg: at the end of the day, kibble - currently - doesnt give a shit about what those keys/codes are in terms of standards or naming conventions.

We just need to pick a name for the value that is put in urls e.g `/en/`, `/es-mx/`, or `/esp/`, and a name for the value that is used to get language files e.g. `en_EN`, `eng`, or `en-GB`.

I am happy to use the same standard for the latter that html/xml uses, which @stajs suggested.

stajs (approved):

PeculiarGoat:
Language selector styling/templating is done!  I've deviated from the design team mockup a bit just to get it working quickly.  We can make it better when we replace the nav styles/template.

PR #63, opened 2020-11-12
Language file to CSV script

PeculiarGoat:
Outputs an up-to-date language CSV which can be imported into Google Sheets using '~' as the separator.

Filters out language vars that don't have a value yet (subscription and pin code related stuff that we haven't used).

The idea here is to be able to quickly give the most up-to-date list of translations to Todd in a format 90% of the way to being presentable to a new client.  Thoughts @stajs @l0ud0gg?

### DIFF
--- a/scripts/language-file-to-csv/language-file-to-csv.js
+++ b/scripts/language-file-to-csv/language-file-to-csv.js
@@ -29,13 +29,13 @@ for (const [key, value] of Object.entries(json)) {
 
 // Because this character '~' is not present in the English language file.
 // May have to change how this works in the future.
-let csvSeparator = '~'; 
+let csvSeparator = '","'; 
 
 let createRow = (key, value, singularValue) => {
-  return [key, '', value, '', singularValue].join(csvSeparator);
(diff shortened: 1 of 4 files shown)


stajs (approved):
Looks great! Got an example CSV screenshot to show how it handles `one` and `other`s?

scripts/language-file-to-csv/language-file-to-csv.js
```diff
@@ -0,0 +1,55 @@
...
+for (const [key, value] of Object.entries(json)) {
+  if (key === value.other) {
+    console.warn(`WARNING - Unused key '${key}' will be removed!`);
+    delete json[key];
+  }
+}
+
+// Because this character '~' is not present in the English language file.
+// May have to change how this works in the future.
+let csvSeparator = '~';
```
gscragg, line 32: This seems a little dangerous. Perhaps there is some easy js you could use that makes sure strings are quoted and escaped properly. At the very least:
```js
let field = `"${value.replace(/"/g, '""')}"`
```

PeculiarGoat:
@stajs @gscragg I just added a language file validator script to run some basic (and janky) tests on the language files.  Could possibly set up Drone for this repo?

gscragg (approved):

stajs (approved):

PR #67, opened 2020-11-20
Adjust font size of disabled text fields

additiverich:
Don't increase the font size or padding of read-only fields when they are text fields (not masked "****" fields)

This is with reference to when dates of birth aren't editable

Screen Shot 2020-11-20 at 3 37 16 PM

### DIFF
--- a/site/styles/_shift72.scss
+++ b/site/styles/_shift72.scss
@@ -135,9 +135,11 @@
 
 // some fields which are "readonly" shouldn't have the padding and font-size applied to them because they are text (not masked "****" values)
 .s72-readonly-text .s72-form-control-readonly {
+  @extend .py-0;
+
   font-size: unset;
-  padding: 0 0.875rem;
-  border-radius: 0.25rem !important;
+  border-radius: $border-radius !important;
+  border: 1px solid transparent;
 }
 
 s72-useraccount-form {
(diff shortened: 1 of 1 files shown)


stajs (approved):
All good, a couple of (unrelated to your change) things:

Any chance we can nudge the readonly text over a pixel or two? Not even looking at the damn `select`
image

Also, the corner radius doesn't look the same as editable fields.

additiverich:
@stajs Well spotted. Changed the padding left to 14px and border radius to 4px
Screen Shot 2020-11-22 at 9 07 21 PM

site/styles/_shift72.scss
```diff
@@ -133,6 +133,13 @@
   }
 }
 
+// some fields which are "readonly" shouldn't have the padding and font-size applied to them because they are text (not masked "****" values)
+.s72-readonly-text .s72-form-control-readonly {
+  font-size: unset;
+  padding: 0 0.875rem;
+  border-radius: 0.25rem !important;
```
gscragg, line 140: These values look like hacks.
Why not use the variables that exist for site wide [border-radius](https://github.com/twbs/bootstrap/blob/main/scss/_variables.scss#L654) and [padding](https://github.com/twbs/bootstrap/blob/main/scss/_variables.scss#L630)?

gscragg (approved):

PR #68, opened 2020-11-29
Add code for icons-512

smitsanghvi21:
Included code for icons-512 in the manifest file which would help us to not copy/paste into a new site each time.


PeculiarGoat (approved):
Nice!

PR #69, opened 2020-11-29
Sticky footer v2

curtisadams28:
**Round 2**

This method makes the min-height of the page = view height - footer height. The footer height is determined by a variable that can be changed depending on the image size.

ADO: https://dev.azure.com/S72/SHIFT72/_workitems/edit/1436
Sticky footer v1: https://github.com/shift72/core-template/pull/52


stajs:
Sorry, I like the first flexbox PR better - I don't want to have another variable to adjust the footer height if we can avoid it.

https://philipwalton.github.io/solved-by-flexbox/demos/sticky-footer/

PR #71, opened 2020-12-01
ie11 password reset justify bug fix

PeculiarGoat:
I can't actually test that this works (Ubuntu)... would one of you mind checking that it's okay before I merge please.


stajs:
@curtisadams28 I'm not familiar with the bug, are you? Can I leave with you to test?

curtisadams28:
It works but wouldn't it be better just to remove the `justify-content-center` class entirely? Searching through core, it is only used in forgotpassword.jet

PeculiarGoat:
@curtisadams28 probably yeah

PR #72, opened 2020-12-01
add margin to form text on signup page

PeculiarGoat:


smitsanghvi21 (approved):
Works fine on all the devices, scales on zoom in and out. Just a tiny thing which isn't important though-bit of extra spacing above the text if there's a Newsletter checkbox.
image

stajs:
What are we fixing and why? The ADO card has no detail on it. The title "create account button padding is wrong" - what is wrong? Before and after pics? We need to be better with keeping the cards clear about what the work is.

PeculiarGoat:
@stajs sorry I forgot to fill it out... it's done now: https://dev.azure.com/S72/SHIFT72/_workitems/edit/2481

PR #75, opened 2020-12-04
Ab#1466 modal issues

PeculiarGoat:
[Two bugs on one card](https://dev.azure.com/S72/SHIFT72/_workitems/edit/1466):
1. I still can't figure out what is allowing pages to scroll behind modals (and it's not just on iPhone or the account page).  Fixing this might be even more difficult while [this bug](https://dev.azure.com/S72/SHIFT72/_workitems/edit/2530) exists.
2. Fixed it.


smitsanghvi21:
@PeculiarGoat Still can't see the Footer when the modal of the account page is open though (screen recording [here](https://monosnap.com/file/5K3HeuKqd51J26ph8UkONFwiJd4dvH)). However, looks like the 1st issue might be fixed because I wasn't able to replicate it when cloned shift72-template (screenshot [here](https://monosnap.com/file/fnB7JBYxsbfSWH2EfnYanlIRTN5kmt)).

PeculiarGoat:
@smitsanghvi21 I think the footer on the account page is actually below the screen!

gscragg (approved):

PeculiarGoat:
I can't find any side-effects to removing `overflow-y: scroll` from the html element to fix the scrolling problem.  Might be worth getting this tested before merging anyway though.

PR #76, opened 2020-12-07
Bonus content fix

curtisadams28:
https://dev.azure.com/S72/SHIFT72/_boards/board/t/Delivery%20team/Stories/?workitem=540

### DIFF
--- a/site/styles/_meta-detail.scss
+++ b/site/styles/_meta-detail.scss
@@ -397,7 +397,6 @@
               font-size: 12px;
             }
           }
-
           .runtime {
             margin-left: auto;
             font-size: 8px;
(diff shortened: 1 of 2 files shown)


PeculiarGoat (changes requested):
Your solution works nicely but I don't like the visual inconsistency that having a long/short synopsis creates with the runtimes.
image

I think you should run this past the design team and let them decide if we need to change the way the runtime is aligned or if we can live how it is in the screenshot.

curtisadams28:
png)
> 
> I think you should run this past the design team and let them decide if we need to change the way the runtime is aligned or if we can live how it is in the screenshot.

PR #81, opened 2020-12-22
remove offset height from padding

smitsanghvi21:
Based on [this](https://dev.azure.com/S72/SHIFT72/_workitems/edit/2655/) issue. Looks like the overflow was happening due to offset height being added in the subnavOverflowing function. This might not be required anymore since couple of weeks ago the default padding was adjusted in the variable file.

### DIFF
--- a/site/static/js/main.js
+++ b/site/static/js/main.js
@@ -489,7 +489,7 @@ function togglePageTopPadding() {
   if (subnavOverflowing()) {
     if (page.hasAttribute('style')) return;
     var padding = parseFloat(window.getComputedStyle(page, null).getPropertyValue('padding-top'));
-    page.style.paddingTop = (padding + navItem.offsetHeight) + 'px';
+    page.style.paddingTop = (padding) + 'px';
   } else {
     page.removeAttribute('style');
   }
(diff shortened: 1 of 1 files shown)


PeculiarGoat:
Adjusting the padding in the variable file wont have an effect on subnav wrapping.  Does removing the offset height prevent stop the page from bumping down when the subnav wraps?... if yes... it might be better to just remove the subnavOverflowing function completely and rethink how to solve the subnav wrapping issue.

smitsanghvi21:
@PeculiarGoat If i am not wrong this would just add some padding to the page class which is below the navigation when the subnav overflows right? Offset height is just adding some extra padding between the nav and the page content while causing this issue when subnav overflows. Removing it fixes the issue and reduced the padding but still leaves some padding between those 2 divs and doesn't mess the layout.

PeculiarGoat:
I just tested it out.... the `togglePageTopPadding` is made redundant by your change so I think it should just be removed.  The padding on the carousel stuff has been increased recently (I think) so the subnav wrapping shouldn't be an issue anymore unless the client wants a crazy amount of subnav items like we have on the staging site (it wraps to 4 rows :-1:).

PeculiarGoat:
@smitsanghvi21 where's this at?

PeculiarGoat (approved):
:+1:

PR #83, opened 2021-01-10
full width carousel background

curtisadams28:
ADO Card (https://dev.azure.com/S72/SHIFT72/_boards/board/t/Delivery%20team/Stories/?workitem=2706)

### DIFF
--- a/site/templates/collection/carousel_item.jet
+++ b/site/templates/collection/carousel_item.jet
@@ -13,7 +13,7 @@
     <a href="{{routeToSlug(item.Slug)}}" class="carousel-item-link">
 
       {{if isset(item.Images["Carousel"])}}
-        <s72-image src="{{item.Images.Carousel}}" alt="{{item.Title}}" class="carousel-item-image"></s72-image>
+        <div class="carousel-item-image" style="background-image: url({{item.Images.Carousel}})"></div>
       {{end}}
 
       {{if !isset(item["PageType"]) || item.PageType != "external"}}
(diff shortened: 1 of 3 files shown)


site/templates/collection/carousel_item.jet
```diff
@@ -13,7 +13,7 @@
     <a href="{{routeToSlug(item.Slug)}}" class="carousel-item-link">
 
       {{if isset(item.Images["Carousel"])}}
-        <s72-image src="{{item.Images.Carousel}}" alt="{{item.Title}}" class="carousel-item-image"></s72-image>
+        <div class="carousel-item-image" style="background-image: url({{item.Images.Carousel}})"></div>
```
curtisadams28, line 16: Ideally, I would have used `object-fit: cover;` on the image but it isn't supported in IE11. The tradeoff with this method is that you don't get the alt tag. Alternatively, I discovered that you can use `background-image` on a `` tag but I'm not sure if that is good practice?

gscragg (changes requested):
I would like a good reason why `s72-image` is no longer being used.

gscragg:
> I would like a good reason why `s72-image` is no longer being used.

And I ask this because using an `s72-image` component gives us the power to change the image that is rendered or make other decisions about the image before rendering it.

Our carousel component does try to resize itself based on the images dimensions: https://github.com/indiereign/shift72-web/blob/master/src/ui/carousel/carousel.component.js#L10

stajs:
@gscragg New PR here https://github.com/shift72/core-template/pull/86

@curtisadams28 This PR should be abandoned? The carousel `imageSizeRatio` - is that what you've done in the CSS calcs of the new PR (maybe continue discussion there)?

PR #84, opened 2021-01-11
Look for a specific value of 'wishlist_enabled' before displaying nav…

additiverich:
… item

(if the value has been set to something other than "true" it'll still behave as if it's true; the only way this can not be true is if it does not exist)


PeculiarGoat (approved):
Heh I noticed this last week too!

Might be worth locking the component and button behind the config in Relish too.

PR #86, opened 2021-01-12
Full width carousel v2

curtisadams28:
This method is a lot cleaner but uses `@media`. Like the current method, it assumes that the image will be 1600 x 600. You can see the differences between the methods here:
https://shift72.slack.com/files/U0183MW9TNK/F01K5STE3FA/background-matrix-v2.jpg

### DIFF
--- a/site/styles/_carousel.scss
+++ b/site/styles/_carousel.scss
@@ -22,6 +22,8 @@ s72-carousel {
   s72-image {
     width: 100%;
     height: $carousel-height;
+    position: relative;
+    overflow: hidden;
     @include media-breakpoint-up(sm) {
       height: $carousel-height-sm;
     }
--- a/site/styles/_carousel.scss
+++ b/site/styles/_carousel.scss
@@ -22,10 +22,6 @@ s72-carousel {
   s72-image {
     width: 100%;
     height: $carousel-height;
-    max-width: 1600px;
-    position: relative;
-    float: right;
-    overflow: hidden;
     @include media-breakpoint-up(sm) {
       height: $carousel-height-sm;
     }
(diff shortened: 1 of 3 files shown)


site/styles/_carousel.scss
```diff
@@ -50,17 +48,24 @@ s72-carousel {
...
+      transform: translate(-50%, -50%);
       @include media-breakpoint-up(sm) {
         height: $carousel-height-sm;
       }
       @include media-breakpoint-up(md) {
         width: auto;
         height: $carousel-height-md;
       }
+      // Finds the width of the image given the carousel height and a fixed aspect ratio of 1600:600
+      $carousel-image-width: (1600 / 600) * $carousel-height-md;
```
PeculiarGoat, line 64: Does this work at the default and sm breakpoints too?

curtisadams28: Yes. `$carousel-image-width` is used to detect when the screen width is greater than the image width. This isn't going to happen with small screen widths.

stajs: Is this doing the same as `imageSizeRatio` [in the component](https://github.com/indiereign/shift72-web/blob/master/src/ui/carousel/carousel.component.js#L10)?

gscragg: I think that ratio should be variable(s) so if a site has different carousel image size this code should still be reasonably happy.

I also think someone (me, @PeculiarGoat, anyone) should add an attribute to the carousel component to disable resizing which might make these changes a little easier to wrangle.

curtisadams28: @stajs No. That code is calculating the height of `.s72-carousel-slides` from the image width. When the image width changes, the height style changes in `.s72-carousel-slides`.  I don't think it's being used though as it is overwritten by the `carousel-height` variables.
image

curtisadams28: @gscragg Agreed. I'll add it in.

curtisadams28:
Something @stajs has mentioned is some existing clients won't want their images cut off at the top and bottom by this new method. I'm thinking it might be worth having this as a mixin so that we can toggle between the new and old version. It will make things easier when we go back to old sites to update them.

curtisadams28:
Closing as we are trying a different approach.

PR #87, opened 2021-01-15
changed cursor on language selector

curtisadams28:
Just changed the cursor to be consistent with the other dropdown menus.


PeculiarGoat (approved):
Cool I didn't notice that!

PR #94, opened 2021-04-16
New nav

curtisadams28:
This is early on and it needs a lot of changes. There are a lot of things hardcoded that shouldn't be as I haven't created any variables yet.


site/templates/nav.jet
```diff
@@ -2,73 +2,67 @@
...
+
       {{if config("search_disabled") != "true"}}
-        <form class="form-search navbar-nav-search" action="{{ routeToPath("/search.html") }}">
-            <input class="form-control form-control-search" type="search" placeholder="{{i18n("search_control_placeholder")}}" name="q" aria-label="{{i18n("search_control_placeholder")}}">
-            <button class="sr-only" type="submit">{{i18n("search_control_submit")}}</button>
+        <form class="form-search navbar-nav-search search-desktop"  action="{{ routeToPath("/search.html") }}">
+          <button type="button" class="search-open" onclick="focusSearch()">
+            <svg class="search-icon" width="24" height="24" viewBox="0 0 24 24"><path d="M9.5,3A6.5,6.5 0 0,1 16,9.5C16,11.11 15.41,12.59 14.44,13.73L14.71,14H15.5L20.5,19L19,20.5L14,15.5V14.71L13.73,14.44C12.59,15.41 11.11,16 9.5,16A6.5,6.5 0 0,1 3,9.5A6.5,6.5 0 0,1 9.5,3M9.5,5C7,5 5,7 5,9.5C5,12 7,14 9.5,14C12,14 14,12 14,9.5C14,7 12,5 9.5,5Z" /></svg>
+          </button>
+          <input class="form-control form-control-search" type="search" placeholder="Search" onfocusout="focusOutSearch()" name="q" aria-label="{{i18n("search_control_placeholder")}}">
```
PeculiarGoat, line 18: Calling js functions directly on elements is not going to work for shift72-template sites because of the way rollup.js smushes all the js files together.  It would be better to setup an [event listener](https://developer.mozilla.org/en-US/docs/Web/API/Element/focusout_event) directly in main.js.

site/templates/nav.jet
```diff
@@ -2,73 +2,67 @@
...
+          <button class="search-close">
+            <svg xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" version="1.1" width="24" height="24" viewBox="0 0 24 24"><path d="M19,6.41L17.59,5L12,10.59L6.41,5L5,6.41L10.59,12L5,17.59L6.41,19L12,13.41L17.59,19L19,17.59L13.41,12L19,6.41Z" /></svg>
+          </button>
+          <button class="sr-only" type="submit">{{i18n("search_control_submit")}}</button>
         </form>
       {{end}}
 
-      <button class="navbar-toggler" type="button" toggle-function="toggleMobileMenu" s72-collapse=".navbar-collapse" aria-controls="navbarsExampleDefault" aria-expanded="false" aria-label="Toggle navigation">
-        <span class="icon-navbar-toggler"></span>
+      <button class="navbar-toggler" type="button" onClick="toggleHidden()" toggle-function="toggleMobileMenu" s72-collapse=".navigation" aria-controls="navbarsExampleDefault" aria-expanded="false" aria-label="Toggle navigation">
```
PeculiarGoat, line 26: Same issue as the other comment about calling functions on elements.  Also... I'm pretty sure `toggle-function="toggleMobileMenu"` has been broken since we added rollup.js to shift72-template so probably a good idea to look at alternative solutions.

site/templates/nav.jet
```diff
@@ -2,73 +2,67 @@
...
 
 {{block nav(showSubNav=true)}}
-<div class="navigation" data-toggle="affix">
-  <nav class="navbar navbar-expand-md navbar-dark">
-      <a class="navbar-brand" href="{{ routeToPath("/") }}">{{i18n("nav_homepage")}}</a>
-      
+<div class="navigation search-hidden" data-toggle="affix" aria-hidden="true">
+  <nav>
+
+    <div class="navbar">
```
PeculiarGoat, line 8: Isn't this `div` redundant?  Why not do `<nav class="navbar">` like it was before?

site/templates/nav.jet
```diff
@@ -2,73 +2,67 @@
...
+        <a class="navbar-brand" href="{{ routeToPath("/") }}">{{i18n("nav_homepage")}}</a>
+      </div>
+
       {{if config("search_disabled") != "true"}}
-        <form class="form-search navbar-nav-search" action="{{ routeToPath("/search.html") }}">
-            <input class="form-control form-control-search" type="search" placeholder="{{i18n("search_control_placeholder")}}" name="q" aria-label="{{i18n("search_control_placeholder")}}">
-            <button class="sr-only" type="submit">{{i18n("search_control_submit")}}</button>
+        <form class="form-search navbar-nav-search search-desktop"  action="{{ routeToPath("/search.html") }}">
+          <button type="button" class="search-open" onclick="focusSearch()">
+            <svg class="search-icon" width="24" height="24" viewBox="0 0 24 24"><path d="M9.5,3A6.5,6.5 0 0,1 16,9.5C16,11.11 15.41,12.59 14.44,13.73L14.71,14H15.5L20.5,19L19,20.5L14,15.5V14.71L13.73,14.44C12.59,15.41 11.11,16 9.5,16A6.5,6.5 0 0,1 3,9.5A6.5,6.5 0 0,1 9.5,3M9.5,5C7,5 5,7 5,9.5C5,12 7,14 9.5,14C12,14 14,12 14,9.5C14,7 12,5 9.5,5Z" /></svg>
```
PeculiarGoat, line 16: We use Font Awesome 4 for icons... no need for svg chaos!

site/templates/nav.jet
```diff
@@ -2,73 +2,67 @@
...
+          </button>
+          <input class="form-control form-control-search" type="search" placeholder="Search" onfocusout="focusOutSearch()" name="q" aria-label="{{i18n("search_control_placeholder")}}">
+          <button class="search-close">
+            <svg xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" version="1.1" width="24" height="24" viewBox="0 0 24 24"><path d="M19,6.41L17.59,5L12,10.59L6.41,5L5,6.41L10.59,12L5,17.59L6.41,19L12,13.41L17.59,19L19,17.59L13.41,12L19,6.41Z" /></svg>
```
PeculiarGoat, line 20: We use Font Awesome 4 for icons... no need for svg chaos!

site/templates/nav.jet
```diff
@@ -2,73 +2,67 @@
...
+            {{if len(.Items) > 0}}
+              <li class="nav-item">
+                <s72-dropdown>
+                  <a class="s72-dropdown-toggle nav-link" href="{{len(.Link.ExternalURL) > 0 ? .Link.ExternalURL : len(.Items) == 0 ? routeToSlug(.Link.Slug) : "#"}}">{{.Label}}</a>
+                  <div class="s72-dropdown-menu sub-nav-menu">
+                    {{range sub := .Items}}
+                      <a class="nav-link" href="{{len(sub.Link.ExternalURL) > 0 ? sub.Link.ExternalURL : routeToSlug(sub.Link.Slug) }}">{{sub.Label}}</a>
+                    {{end}}
+                  </div>
+                  <svg  width="24" height="24" viewBox="0 0 24 24" class="dropdown-caret"><path d="M7.41,8.58L12,13.17L16.59,8.58L18,10L12,16L6,10L7.41,8.58Z" /></svg>
```
PeculiarGoat, line 52: We use Font Awesome 4 for icons... no need for svg chaos!

site/templates/nav/user_logged_in.jet
```diff
@@ -3,10 +3,15 @@
     <s72-dropdown class="navbar-nav-user-nav">
       <div>
         <a href="#" class="s72-btn s72-dropdown-toggle">
-          {{i18n("nav_signed_in_as")}} <s72-username></s72-username>
+          <svg version="1.1" id="Layer_1" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" class="user-icon" width="24" height="24" x="0px" y="0px" viewBox="-243 245 24 24" xml:space="preserve">
```
PeculiarGoat, line 6: We use Font Awesome 4 for icons... no need for svg chaos!

kibble.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "core-template",
   "version": "0.0.1",
-  "siteUrl": "https://staging-abccinemas.screenplus.co",
+  "siteUrl": "https://abccinemas.screenplus.co",
```
PeculiarGoat, line 4: Better remove this change! :eyes:

PeculiarGoat (changes requested):
Why did you change the abc-cinemas logos to Shift72 logos?

PeculiarGoat (changes requested):
Not sure that we want all those new pngs/svgs in the icons folder.  Font Awesome should have us covered.

site/styles/_variables.scss
```diff
@@ -33,10 +33,11 @@ $container-max-widths: (
 // Body / Core
 // ------------------------------------
 
-$primary:         #35a0a8 !default;
-$secondary:       #4cd3dd !default;
-$body-bg:         #000 !default;
-$body-color:      #fff !default;
+$primary:         #4d97ff !default;
```
PeculiarGoat, line 36: I don't think we need to anything changed/added in this file.

site/styles/_variables.scss
```diff
@@ -33,10 +33,11 @@ $container-max-widths: (
...
+$secondary:       #2d85ff !default;
+$body-bg:         #0F0F0F !default;
+$body-bg-light:   #181818 !default;
```
PeculiarGoat, line 39: You can lighten the existing bg colour instead adding another var.

site/templates/nav.jet
```diff
@@ -2,73 +2,67 @@
...
+            <svg xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" version="1.1" width="24" height="24" viewBox="0 0 24 24"><path d="M19,6.41L17.59,5L12,10.59L6.41,5L5,6.41L10.59,12L5,17.59L6.41,19L12,13.41L17.59,19L19,17.59L13.41,12L19,6.41Z" /></svg>
+          </button>
+          <button class="sr-only" type="submit">{{i18n("search_control_submit")}}</button>
         </form>
       {{end}}
 
-      <button class="navbar-toggler" type="button" toggle-function="toggleMobileMenu" s72-collapse=".navbar-collapse" aria-controls="navbarsExampleDefault" aria-expanded="false" aria-label="Toggle navigation">
-        <span class="icon-navbar-toggler"></span>
+      <button class="navbar-toggler" type="button" onClick="toggleHidden()" toggle-function="toggleMobileMenu" s72-collapse=".navigation" aria-controls="navbarsExampleDefault" aria-expanded="false" aria-label="Toggle navigation">
+        <span class="hamburger-top"></span>
```
PeculiarGoat, line 27: Is this a homemade :hamburger: ?  Font Awesome please!

PeculiarGoat:
It's a solid start... you've removed that horrible duplicated subnav!  I think you shouldn't worry about new styles though... just focus on recreating the appearance and functionality of the old nav with dev-friendly html/css.  Once that is sorted... then we can look at the new styles and it should be super easy to work on.

site/templates/nav.jet
curtisadams28: Design specifically wanted icons from here: (https://materialdesignicons.com) and to be fair, the Font Awesome icons look pretty terrible. I do agree that inserting them as SVG's is not the greatest though. Do you know if we have the pro version of font awesome? Because then we could upload any icons we want. Alternativley, I could add them as blocks in a .jet file and use `yield` to call them. That would tidy it up.

site/templates/nav.jet
curtisadams28: I needed to make the burger because I've animated it.

site/templates/nav.jet
PeculiarGoat: Hmm okay... then I think that we should have a separate task to change all icons across the site (might need some Relish work too) instead of changing a few for the nav task.  And I'm pretty sure that MDI icons just need classes like Font Awesome... no SVG lol.

curtisadams28: Replaced with Font Awesome

site/templates/nav.jet
curtisadams28: Replaced with Font Awesome

site/templates/nav/user_logged_in.jet
curtisadams28: Replaced with Font Awesome

site/templates/nav.jet
curtisadams28: Fixed.

curtisadams28:
> Why did you change the abc-cinemas logos to Shift72 logos?

The Shift72 logo was in the design

site/templates/nav.jet
curtisadams28: fixed

site/templates/nav.jet
curtisadams28: fixed

site/templates/nav.jet
curtisadams28: Fixed. I set the event listeners in main.js

site/styles/_variables.scss
curtisadams28: This is the new colour that design wants to have as default.

curtisadams28:
> It's a solid start... you've removed that horrible duplicated subnav! I think you shouldn't worry about new styles though... just focus on recreating the appearance and functionality of the old nav with dev-friendly html/css. Once that is sorted... then we can look at the new styles and it should be super easy to work on.

The styles are already mostly done.

smitsanghvi21 (changes requested):
Looks good, couple of comments:

- Sub-menus on mobile devices don't seem to be having a drop-down icon which might indicate that there's nothing inside it. Would be ideal to have an icon just like we have for desktop. 
- Padding of Navbar doesn't seem to be aligned with that of Footer anymore.
- On click events being logged in console.
- Site logo doesn't route to homepage for mobile devices.
- Not sure what does the main div `navigation search-hidden` does as doesn't seem to be hiding for any breakpoint or is rendered in a config.
- The search bar doesn't seem to be working for Safari on MacOS. Also doesn't seem to work for iOS devices when testing using Browserstack. [Link here](https://monosnap.com/file/omOMgA6iOPNCh8tgImhUk8eP06cbdB)

PeculiarGoat:
@curtisadams28 probably a good idea to merge master into this branch and resolve the merge conflicts before going any further!

PeculiarGoat:
@curtisadams28 I think if you change the PR merge target from master to develop, it should cut down the number of files changes and make this easier to review!

site/styles/_nav.scss
```diff
@@ -1,77 +1,181 @@
...
+// Desktop Order
+@include media-breakpoint-up(md) {
+  .navbar-nav { order: 1; }
+  .sub-nav { order: 2; }
+}
+
+/*** HAMBURGER MENU TOGGLER ***/
+
+.navbar-toggler {
+  position: absolute;
```
PeculiarGoat, line 104: Absolute positions were the main reason for re-doing the nav in the first place... really not keen on having more absolute things now.

I think you will be able to achieve the same result visually using [grid layout](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Grid_Layout).  Looks like grid doesn't have good IE11 support... are we allowed to use it @stajs ?

Here is an html file I slapped together to test it out:
```
<!DOCTYPE html>
<html>
  <head>
    <style type="text/css">
      .wrapper {
        display: grid;

        /*
          Three columns:
            - First is max width it can be.
            - Second and third are the minimum width they can be.
        */
        grid-template-columns: auto repeat(2, min-content);

        /*
          Four rows:
            - All minimum height they can be.
        */
        grid-template-rows: repeat(4, min-content);
      }
      .brand {
        grid-column: 1;
        grid-row: 1;
      }
      .search {
        grid-column: 2;
        grid-row: 1;
      }
      .toggler {
        grid-column: 3;
        grid-row: 1;
      }
      .subnav {
         /* Full width of wrapper. */
        grid-column: 1 / -1;
        grid-row: 3;
      }
      .language {
         /* Full width of wrapper. */
        grid-column: 1 / -1;
        grid-row: 2;
      }
      .account {
         /* Full width of wrapper. */
        grid-column: 1 / -1;
        grid-row: 4;
      }
      @media screen and (min-width: 768px) {
        .wrapper {
          grid-template-columns: auto repeat(3, min-content);
          grid-template-rows: repeat(2, min-content);
        }
        .toggler {
          display: none;
        }
        .subnav {
          grid-row: 2;
        }
        .language {
          grid-column: 3;
          grid-row: 1;
        }
        .account {
          grid-column: 4;
          grid-row: 1;
        }
      }

      .brand {
        background-color: red;
      }
      .search {
        background-color: blue;
      }
      .toggler {
        background-color: teal;
      }
      .subnav {
        background-color: green;
      }
      .language {
        background-color: blueviolet;
      }
      .account {
        background-color: darkgoldenrod;
      }
    </style>
  </head>
  <body>
    <header>
      <div class="wrapper">
        <div class="brand">BRAND</div>
  
        <div class="search">SEARCH</div>
  
        <div class="toggler">TOGGLER</div>
  
        <nav class="subnav">
          <ul>
            <li>SUBNAV#1</li>
            <li>SUBNAV#2</li>
            <li>SUBNAV#3</li>
          </ul>
        </nav>
  
        <div class="language">LANGUAGE</div>
  
        <div class="account">
          <span>SIGN IN</span>  
          <span>SIGN UP</span>
        </div>
      </div>
    </header>
  </body>
</html>
```

stajs: You have to get core/web teams on board to drop IE11 support first.

Perhaps it's worth seeing how far you can get with the limitations of using the `-ms-` vendor prefix?

site/styles/_variables.scss
```diff
@@ -37,6 +37,7 @@ $primary:         #35a0a8 !default;
 $secondary:       #4cd3dd !default;
 $body-bg:         #0F0F0F !default;
 $body-color:      #fff !default;
+$body-bg-light:   lighten($body-bg, 4%) !default;
```
stajs, line 40: How important is this, and what if the bg is white?

site/styles/_variables.scss
```diff
@@ -89,7 +87,8 @@ $signin-text-color: $body-color !default;
 $signin-text-color-hover: $body-color !default;
 $signin-border-color: transparent !default;
 $signin-border-color-hover: transparent !default;
-$signin-background: transparent !default;
+$signin-background: rgba($color: $body-color, $alpha: 0.06) !default;
```
stajs, line 90: That sure is the tiniest alpha - does it work across light/dark themes?

site/styles/main.scss
```diff
@@ -357,3 +357,10 @@ s72-classification-label {
 h1, h2, h3, h4, h5, h6, p, .title {
   word-wrap: break-word;
 }
+
+.noscroll {
```
stajs, line 361:
```suggestion
.no-scroll {
```

site/styles/main.scss
```diff
@@ -357,3 +357,10 @@ s72-classification-label {
...
+  overflow: hidden;
+  @include media-breakpoint-up(md) {
```
stajs, line 363: Why the different behaviour?

site/styles/_nav.scss
PeculiarGoat: Also... IE11 _does_ have partial support for grid so maybe it will be good enough to do what we want.  Need to do some more investigation.

site/templates/nav/user_logged_out.jet
```diff
@@ -1,5 +1,6 @@
 {{block nav_user_logged_out()}}
   <s72-user-anon class="user-anon">
+    <div class="vl"></div>
```
stajs, line 3: What's this about?

stajs (commented):
Some questions.

curtisadams28:
@PeculiarGoat @stajs I didn't think css grid was an option. The issue I've been having is that if you follow the designs for mobile, the toggled menu should be on top of the brand and search. 

Mobile Menu

You can fake it by hiding the brand and search when the menu is toggled but this only works if it isn't animated. If you were to add a slide in animation for example, you would see the brand and search disappear as the menu moved. The brand/search and the rest of the menu behave like two separate containers whereas the desktop is just one. You can see this best from the tablet view where you can no longer just hide the brand.

Two Containers

I was trying to work around this issue with absolute positioning and moving things outside the nav container. It just made it more hax though and didn't fully solve it.

Probably the easiest way to solve this is to not have animations and to have an absolutely positioned brand outside the container. BUT if css grid can solve this, that would be great. Thoughts?

In regards to grid vs flexbox, flexbox can actually achieve the same thing. If the brand and search are hidden, then we have a simple column of elements. That is how I had it originally before I started messing with animations.

Mobile Flex

stajs:
@curtisadams28 Thanks for the extra info. I haven't been considering the designs at all here, just code reviewing. I would need to set aside a block of time to dig in to this to really wrap my head around the problem/solution, but it's not likely soon :/

site/en_AU.all.json
```diff
@@ -44,8 +44,8 @@
 
 
   "nav_homepage": { "other": "ABC Cinemas" },
-  "nav_signin": { "other": "Sign In" },
-  "nav_signup": { "other": "Create Account" },
+  "nav_signin": { "other": "Log In" },
+  "nav_signup": { "other": "Sign Up" },
```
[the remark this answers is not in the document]
curtisadams28: Yes. I have changed it back.

site/styles/main.scss
curtisadams28: Done

site/styles/main.scss
curtisadams28: So if the user resizes the window back to desktop, they can scroll.

site/templates/nav/user_logged_out.jet
curtisadams28: It's a vertical line that seperates the login/logout from the search and language selector. It is purely visual. I would use a border for this but in the designs, it is shorter than the height of the container. As far as I'm aware, borders can only be the full height.

PeculiarGoat:
:+1: for ditching absolute positioning!

I'm still not keen on this html structure... I don't like having multiple brand logos and I don't like everything else being inside a nav landmark.

Also, keyboard control is busted on that fancy search bar!

site/styles/_nav.scss
PeculiarGoat: IMO css grid is viable judging from this: https://dev.to/eblairmckee/css-grid-in-ie11-it-s-possible-and-not-as-hard-as-you-think-okp

site/styles/_variables.scss
curtisadams28: Yes. It's a background colour so the contrast doesn't need to be very high to be noticable.

site/styles/_variables.scss
curtisadams28: @stajs This is needed because there is a new background colour for the mobile menu and dropdowns. With the old nav, you couldn't tell those elements apart from the background because they were the same colour.

site/templates/nav/user_logged_out.jet
stajs: Ah. I was confused because I thought it was `v1` like version one, but it's `vl` for vertical line. Perhaps spell it out in full.
```suggestion
    <div class="vertical-line"></div>
```

Remember that we can push back on anything for discussion. In this case you could easily justify making it a border rather than having another HTML element purely for visuals.

stajs (approved):
LGTM but waiting on answers for @PeculiarGoat's comments. Also, would like to see some results from cross browser testing

site/templates/nav/user_logged_in.jet
curtisadams28: @PeculiarGoat I tried both of these and they didn't work:
https://fontawesome.com/v4.7/icon/user-circle-o
https://fontawesome.com/v4.7/icon/user-circle

This one does but looks bad:
https://fontawesome.com/v4.7/icon/user

Can we stick with the svg temporarily until we get material design icons working?

site/styles/_nav.scss
curtisadams28: Because the nav is practically done, we should save moving it to grid for a version 3.0.

site/en_AU.all.json
```diff
@@ -89,7 +89,7 @@
   "signup_form_dob": { "other": "Birth date" },
   "signup_form_submit": { "other": "Submit" },
 
-  "signin_page_header": { "other": "Sign In" },
+  "signin_page_header": { "other": "Log In" },
```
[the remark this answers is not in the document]
curtisadams28: It has always been Sign In.
image

site/en_AU.all.json
```diff
@@ -107,7 +107,7 @@
   "combined_auth_continue": { "other": "Continue" },
   "combined_auth_change": { "other": "change" },
   "combined_auth_signin": { "other": "Sign in" },
-  "combined_auth_signup": { "other": "Create account" },
+  "combined_auth_signup": { "other": "Sign Up" },
```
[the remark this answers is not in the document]
curtisadams28: done

site/static/js/main.js
```diff
@@ -438,30 +438,39 @@ function cullSlides(slider, direction){
...
-  if (element && icon) {
-    element.classList.contains('show') ? icon.classList.add('toggled') : icon.classList.remove('toggled');
+function navToggle() {
+  var nav = document.querySelector('.navbar-toggler');
+    nav.addEventListener('click', function(e) {
+    noScroll();
+  });
+}
+
+function noScroll() {
```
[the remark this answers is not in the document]
curtisadams28: Done

site/static/js/main.js
```diff
@@ -515,7 +522,10 @@ function documentReady(app) {
   document.querySelectorAll('.navbar-nav').forEach(function(nav) {
     nav.classList.remove('s72-hide');
   });
-
+  if (document.querySelector('.navbar-nav-search')) {
+    searchShow();
```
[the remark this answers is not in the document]
curtisadams28: Done

site/styles/_mixins.scss
```diff
@@ -25,12 +25,36 @@
...
+  color: $link-color;
+  text-decoration: underline;
+  background-color: transparent;
+  &:hover,
+  &:focus {
     color: $link-color;
-    text-decoration: underline;
+    text-decoration: none;
+  }
+}
```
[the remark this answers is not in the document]
curtisadams28: @stajs I merged develop into new-nav so that it was up to date. These changes are in develop.

site/en_AU.all.json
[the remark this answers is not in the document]
stajs: I prefer Sign In. The diff was showing that Sign In was removed and replaced with Log In

image

stajs:
@curtisadams28 Can you push your latest changes for review? Some of the comments you've said "Done" but there are no changes to show in this PR yet.

curtisadams28:
@stajs I have pushed the latest changes.

site/static/js/main.js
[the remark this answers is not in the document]
stajs: Looks like you've renamed the wrong function (than what I was meaning). This function is doing toggle stuff, the other function is setting up the listener and initializing.

site/static/js/main.js
```diff
@@ -438,30 +438,39 @@ function cullSlides(slider, direction){
...
+  searchButton.addEventListener('click', openSearch, false);
+  searchForm.addEventListener('focusin', openSearch, false);
+  searchForm.addEventListener('focusout', closeSearch, false);
 }
 
-function toggleMobileMenu(icon){
-  var element = document.querySelector('.collapse.navbar-collapse');
-  if (element && icon) {
-    element.classList.contains('show') ? icon.classList.add('toggled') : icon.classList.remove('toggled');
+function toggleScroll() {
```
stajs, line 461:
```suggestion
function initToggleScroll() {
```

site/static/js/main.js
```diff
@@ -515,7 +522,10 @@ function documentReady(app) {
   document.querySelectorAll('.navbar-nav').forEach(function(nav) {
     nav.classList.remove('s72-hide');
   });
-
+  if (document.querySelector('.navbar-nav-search')) {
+    initSearch();
+  }
+  toggleScroll();
```
stajs, line 528:
```suggestion
  initToggleScroll();
```

site/static/js/main.js
```diff
@@ -438,30 +438,39 @@ function cullSlides(slider, direction){
...
+  searchForm.addEventListener('focusout', closeSearch, false);
 }
 
-function toggleMobileMenu(icon){
-  var element = document.querySelector('.collapse.navbar-collapse');
-  if (element && icon) {
-    element.classList.contains('show') ? icon.classList.add('toggled') : icon.classList.remove('toggled');
+function toggleScroll() {
+  var nav = document.querySelector('.navbar-toggler');
+    nav.addEventListener('click', function(e) {
```
stajs, line 463:
```suggestion
  nav.addEventListener('click', function(e) {
```

stajs (changes requested):
Just a minor function naming

PR #97, opened 2021-04-27
3965 New Font

cunhapaola:
- changed font family to Poppins -> [ADO card](https://dev.azure.com/S72/SHIFT72/_workitems/edit/3965)

- changed the font-weight based on the [proposed style ](https://zpl.io/bW8k39M)


curtisadams28 (approved):
Looks good! I noticed that some font weights are defined by this variable `$font-weight-bold`. I've kept them in for now as I think we should swap all font weights and sizes to variables in the future.

PR #98, opened 2021-04-29
AB#3878 Giropay

additiverich:
Some changes to support multiple payment options in the shopping session modal [as per designs](https://zpl.io/b6Kmo7m)

Screen Shot 2021-04-30 at 6 53 40 AM

To see these in action, point the site at https://staging-giropay.shift72.com/ (which has CC and giropay as legit payment methods).


curtisadams28 (approved):
Tested on IE, Firefox and Chrome. Looks good to me. I will add those new translation keys to the translations spreadsheet.

PeculiarGoat (commented):
Looks good.  You could add the lang variables to the other lang files :P

PR #100, opened 2021-05-11
fixed film page headings

curtisadams28:
This fixes an issue where the heading rank jumps from h1 to h4. The heading heirarchy can still be broken by the user in the synopsis where thay can define any heading rank.


site/styles/_meta-detail.scss
```diff
@@ -297,7 +297,7 @@
   .meta-detail-country,
   .meta-detail-subtitle {
     padding-bottom: 20px;
-    h3, h4 {
+    h2, h3, h4 {
```
PeculiarGoat, line 300: Did you leave h3 and h4 here for backwards compatibility?

site/styles/_social-media-buttons.scss
```diff
@@ -7,7 +7,7 @@
 .social-media-buttons-title {
 	align-self: center;
 	padding-right: 10px;
-	h5 {
+	h2 {
```
PeculiarGoat, line 10: Should you leave h5 here for backwards compatibility?

PeculiarGoat (changes requested):
I think both css bits should be either backwards compatible or not... not a mixture.

If it's not backwards compatible, this change will break film/social jets on quite a lot of sites... do we just try to remember to fix up each site that gets updated as we do them??  Thoughts @stajs ?

stajs:
Some thoughts...

1. It's quite hard to determine the desired hierarchy based on the PR, it would be a lot easier to understand and discuss as a group if we had an annotated screenshot of all the current headings, and what we think the future state should look like.
2. The intial problem was exacerbated by using headings willy-nilly all over the show and I think some of those choices are a little hard to justify. An `h3`, `h4` and now also a `h2` heading for a subtitle seems to be overstating the importance. How many equivalent levels of each heading are on the page? Perhaps some should be styled without being a `h` tag. Again, having the overall picture of the structure could help the choice here.
3. In terms of compat, what is the size of the problem? If the style and jet change together, then the impact will only be for clients with custom styles/jets - perhaps we could live with that? But I do think we should merge-freeze as many PRs as reasonably possible until NPMifying is done.

site/styles/_meta-detail.scss
curtisadams28: Yes. I'll make sure I make the rest backwards compatable too.

curtisadams28:
Maybe font styles should be included as classes and used in the jet files? I know Bex is going to consolidate font sizes soon so it might be worth revisiting then. For now, I'm going to make things backwards compatible since TIFF needs these changes soon.

curtisadams28:
**Before:**
<h1> Film Title
<h5> Share
<h4> Cast
<h4> Crew
<h4> Languages
<h4> Subtitles
<h4> Countries
<h4> Studio
<h4> Bonus Content
<h3> Recommendations

**After:**
<h1> Film Title
<span> Share
<h2> Cast
<h2> Crew
<h2> Languages
<h2> Subtitles
<h2> Countries
<h2> Studio
<h2> Bonus Content
<h3> Recommendations

stajs:
In terms of heading hierarchy, that makes a _LOT_ more sense. One nit (which I know you didn't change but is now apparent when seeing it laid out): it looks like `Recommendations` is a sub-heading of `Bonus Content`. I think the easiest fix here is to make `Recommendations` a `h2` so it's at a sibling level to the others.

site/styles/_social-media-buttons.scss
curtisadams28: Yup. Have added.

curtisadams28:
Good point @stajs. I've made `Recommendations` an `h2`.

PeculiarGoat:
Nice!

curtisadams28:
@PeculiarGoat How does the latest commit look? I thought it was a bit weird that there's bundle styles in main.scss.

curtisadams28:
**This is also the same for TV shows.**

Before:
`
<h1> Film Title
<span> Share
<h2> Cast
<h2> Crew
<h2> Languages
<h2> Subtitles
<h2> Countries
<h2> Studio
<h2> Bonus Content
   <h5> Individual video heading
<h2> Recommendations
`
After:
`
<h1> Film Title
<span> Share
<h2> Cast
<h2> Crew
<h2> Languages
<h2> Subtitles
<h2> Countries
<h2> Studio
<h2> Bonus Content
   <h3> Individual video heading
<h2> Recommendations
`

site/templates/common/social-media-buttons.jet
```diff
@@ -1,7 +1,7 @@
 {{block socialMediaButtons(path="", title="")}}
 	<div class="social-media-buttons">
 		<div class="social-media-buttons-title">
-			<h5>{{i18n("social_media_buttons_title")}}</h5>
+			<span>{{i18n("social_media_buttons_title")}}</span>
```
PeculiarGoat, line 4: I prefer this as `<h2>`!

PeculiarGoat (changes requested):
Just one minor complaint.  I think the social media buttons header was better as an actual header instead of span.  Some accessibility users will be navigating the page by header so it would be nice if they can get to those buttons quickly and also the heading lets them know when they have reached the end of the main film info section.

site/templates/common/social-media-buttons.jet
curtisadams28: Done :)

PeculiarGoat (approved):
Nice work!  :1st_place_medal:

PR #103, opened 2021-05-20
Ab#4141 missing titles on elements

PeculiarGoat:
https://dev.azure.com/S72/SHIFT72/_workitems/edit/4141 -> **GTM and some Stripe iframes are missing title attributes.**

For the GTM iframe, it is already styled to be invisible (which _should_ be enough for screen readers) but for extra safety I've added aria-hidden and added the title attribute so it will pass automated WCAG testing.

For the Stripe iframes (one appears on DOMContentLoaded, the other appears when a buy/rent button is clicked) and any others without titles, I can't see a way of doing this without a mutation observer to add titles as these iframes can potentially be loaded at different times.  Since the observer is only watching the childlist of body without checking subtree this _should_ have a no performance impact... but it would be possible for a custom jet to be created that adds millions of elements directly under body and hurt performance... can't imagine why anyone would intentionally do this though.

What do you think @stajs?


.npmrc
```diff
@@ -0,0 +1,2 @@
+# Comment out this line to publish to the main registry.
```
stajs, line 1: I'm now thinking this file shouldn't be in git; my mistake I will remove. People who know what they are doing and need to use this can add it.

site/static/js/main.js
```diff
@@ -531,3 +531,15 @@ document.addEventListener('s72loaded', function(event) {
   let app = event.detail.app;
   documentReady(app);
 });
+
+// Last resort iframe title insertion for WCAG compliance.
+new MutationObserver(function(mutations, observer) {
+  mutations.forEach(function(mutation) {
+    mutation.addedNodes.forEach(function(node) {
+      if (node.nodeName !== 'IFRAME') return;
+      if (!node.getAttribute('aria-hidden') || node.getAttribute('title')) return;
```
stajs, line 540: This logic could be at risk if it has `aria-hidden="false"` with a `title` - it will override the given title.

site/google.jet
```diff
@@ -47,7 +47,7 @@
 {{block googleTagManagerNoScript(tagManagerId=config("google_tag_manager_id"))}}
   {{ if len(tagManagerId) > 0 }}
   <!-- Google Tag Manager (noscript) -->
-  <noscript><iframe src="https://www.googletagmanager.com/ns.html?id={{ tagManagerId }}" height="0" width="0" style="display:none;visibility:hidden"></iframe></noscript>
+  <noscript><iframe title="Google Tag Manager" src="https://www.googletagmanager.com/ns.html?id={{ tagManagerId }}" height="0" width="0" style="display:none;visibility:hidden" aria-hidden="true"></iframe></noscript>
```
stajs, line 50: It seems that "[best practice](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques/Using_the_aria-hidden_attribute#best_practices)" is to not use `aria-hidden` if already hidden by styling.

stajs (changes requested):
Can you retarget to the `develop` branch?

Did we confirm the iframes that were triggering the warning? I suspect the GTM one doesn't.

If we have a WCAG compliance issue with Stripe, I'd rather raise an issue with them than add a MutationObserver. If we get pushed hard on it, we can have it in the back pocket, but I think we should have a threshold for potentially negative perf being added as a default to the template.

site/google.jet
PeculiarGoat: Yeah I've seen that... but I've also seen people on Stackoverflow saying that this particular best practice needs to be ignored to get around automated WCAG testing.

site/static/js/main.js
PeculiarGoat: Ah good point!

PR #110, opened 2021-05-24
Ab#4143 missing landmarks on pages

PeculiarGoat:
https://dev.azure.com/S72/SHIFT72/_workitems/edit/4143

Still not a nice experience with the chrome screen reader extension but these landmarks and regions are at least a step towards making the site usable and hopefully will keep TIFF happy.  If you checkout this branch you can have a play around with it and also view the landmark structure with the Wave chrome extension.

TBH I think we might need to bin all of the jets and start over with [inclusive design](http://www.inclusivedesigntoolkit.com/whatis/whatis.html) in mind.  There are random little issues all over core-template that get in the way of screen readers e.g. availability elements coming before film titles, carousels and sliders not working the way you might expect, etc.


site/404.html.jet
```diff
@@ -5,7 +5,7 @@
 {{end}}
 
 {{block body()}}
-<div class="container page not-found">
+<main role="main" class="container page not-found">
```
stajs, line 8: Making the one suggestion here to cover all other instances:

```suggestion
<main class="container page not-found">
```

site/templates/footer.jet
```diff
@@ -1,9 +1,9 @@
 {{block footer()}}
-	<footer>
+	<footer role="contentinfo">
```
stajs, line 2: [Looks to be redundant](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/Contentinfo_role).

```suggestion
	<footer>
```

site/templates/footer.jet
```diff
@@ -1,9 +1,9 @@
...
 		<a href="{{ routeToPath("/") }}" class="footer-brand">
 			{{i18n("nav_homepage")}}
 		</a>
-		<div class="footer-nav">
+		<nav class="footer-nav" role="navigation" aria-label="footer">
```
stajs, line 6: [Looks to be redundant](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/Navigation_Role).
```suggestion
		<nav class="footer-nav" aria-label="footer">
```

site/templates/nav.jet
```diff
@@ -2,14 +2,14 @@
 {{import "./nav/user_logged_in.jet"}}
 
 {{block nav(showSubNav=true)}}
-<div class="navigation" data-toggle="affix">
-  <nav class="navbar navbar-expand-md navbar-dark">
+<header class="navigation" data-toggle="affix" role="banner">
```
stajs, line 5: [Looks to be redundant](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/Banner_role).

```suggestion
<header class="navigation" data-toggle="affix">
```

site/templates/nav.jet
```diff
@@ -2,14 +2,14 @@
...
+  <nav class="navbar navbar-expand-md navbar-dark" role="navigation" aria-label="main">
```
stajs, line 6:
```suggestion
  <nav class="navbar navbar-expand-md navbar-dark" aria-label="main">
```

stajs (changes requested):
Looks like we can save on the redundant bits.

Can I also request that we pick a consistent attribute order, e.g.

`<element role="" class="" other="" />`

PeculiarGoat:
@stajs I've rearranged the attributes now by what I feel is order of importance for the person reading: id -> class -> other stuff.

And those redundant roles aren't actually redundant... different screen readers have different levels of support for landmarks and/or roles.  The Chrome screen reader extension that I'm using (because I can't get Orca to work on Ubuntu) will only acknowledge the top level landmarks (header/main/footer) if they have the role too.  [This site](https://dequeuniversity.com/assets/html/jquery-summit/html5/slides/landmarks.html#best_practice_use_both) claims the best practice is to use both and you can find examples of this in the wild ([bbc.com](https://www.bbc.com/)).

PeculiarGoat:
@stajs I think we should keep the aria roles but you can make the final decision!

site/en_AU.all.json
```diff
@@ -480,5 +480,17 @@
...
+  "shopping_payment_method_header": { "other": "Payment methods" },
+  "shopping_payment_method_credit_card": { "other": "Credit card" },
+  "shopping_payment_method_giropay": { "other": "Giropay" },
+
+  "shopping_error_stripe_unknown_error": { "other": "Something went wrong with this payment. Please try again later." },
+  "shopping_error_payment_complete_failed": { "other": "The payment was declined. Close this window and try again." },
+
+  "accessibility_skip_links_header": { "other": "Accessibility Links" },
+  "accessibility_skip_links_content": { "other": "Skip to Content" },
+  "accessibility_skip_links_footer": { "other": "Skip to Footer" }
```
stajs: I'm sorry, I didn't realise you had already settled on "accessibility" as a prefix in other PRs :/

```suggestion
  "wcag_skip_links_header": { "other": "Accessibility Links" },
  "wcag_skip_links_content": { "other": "Skip to Content" },
  "wcag_skip_links_footer": { "other": "Skip to Footer" }
```

site/templates/application.jet
```diff
@@ -54,6 +54,19 @@
   </head>
   <body>
     <s72-cookie-consent></s72-cookie-consent>
+
+    <section>
```
stajs, line 58: Can we not put the `sr-only` class up here?

stajs (changes requested):
I still think we should drop the aria roles - it feels the same mindset as having IE hacks, we should stick to the standards. WCAG compliance and support for individual screen readers are different levels of work. I understand the state of screen readers on Linux is appalling but I question the statistical viability of that market when they won't be able to play films anyway. I also question the publication date (copyright 2013?) of [that best practice claim](https://dequeuniversity.com/assets/html/jquery-summit/html5/slides/landmarks.html#best_practice_use_both). I found a slightly-less-ancient (2016) [assertion that things are all good](https://www.weba11y.com/blog/2016/04/22/screen-reader-support-for-new-html5-section-elements/) on [Windows/Mac](http://www.html5accessibility.com/) at least.

site/en_AU.all.json
PeculiarGoat: No worries... I prefer wcag_ anyway!

site/templates/application.jet
PeculiarGoat: Yeah.  Weirdly when I tried that before it would still show the <ul> bullet point but it must have just been Chrome flipping out.

PeculiarGoat:
@stajs Chrome's screen reader extension is the issue... not Orca on Linux.  I've removed all of those aria roles anyway (Chrome screen reader doesn't have much market share) except for the search one which is required for the form element to be treated as a landmark.  I've also added translations for the aria labels... completely forgot about it before.

site/templates/nav.jet
```diff
@@ -2,43 +2,46 @@
...
-                  <a class="nav-link" href="{{len(.Link.ExternalURL) > 0 ? .Link.ExternalURL : routeToSlug(.Link.Slug)}}">{{.Label}}</a>
-                </li>
               {{end}}
-            {{end}}
-          </ul>
+            </ul>
         {{end}}
 
-        <div class="navbar-nav s72-hide">
+        <div id="account" class="navbar-nav s72-hide">
```
stajs, line 44: What is targeting this id?

stajs (approved):

PR #114, opened 2021-05-24
Ab#3881 Update language strings for Giropay

l0ud0gg:
Brought curtis changes from master and updated the payment method text in english and german to also include (Germany)


l0ud0gg:
wrong PR

PR #115, opened 2021-05-24
Ab#3881 update language strings for giropay

l0ud0gg:


smitsanghvi21 (approved):
Looks good, do we need to add translations for other languages which comes under Europe for Giropay apart from these ones?

l0ud0gg:
> Looks good, do we need to add translations for other languages which comes under Europe for Giropay apart from these ones?

I guess we will need to add them as we come across them so we have someone to translate for us

PR #116, opened 2021-05-25
remove alt tags from bg images

smitsanghvi21:
Based on [this card](https://dev.azure.com/S72/SHIFT72/_workitems/edit/4140/).

However, the s72-image component isn't just used for the bg images, so I think only needs to be removed where bg images are being used.

### DIFF
--- a/site/templates/bundle/item.jet
+++ b/site/templates/bundle/item.jet
@@ -19,7 +19,7 @@
 
     <div class="meta-detail-bg">
       <div class="meta-detail-bg-gradient"></div>
-      <s72-image src="{{bundle.Images.Background}}" alt="" class="meta-detail-bg-img"></s72-image>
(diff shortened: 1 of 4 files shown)


PeculiarGoat:
From [MDN](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/img#attributes):
> Omitting alt altogether indicates that the image is a key part of the content and no textual equivalent is available. Setting this attribute to an empty string (alt="") indicates that this image is not a key part of the content (it’s decoration or a tracking pixel), and that non-visual browsers may omit it from rendering. Visual browsers will also hide the broken image icon if the alt is empty and the image failed to display.

The attributes that are set on `<s72-image>` are added to the `` inside when the component renders so we actually want to keep these empty alt attributes.

PR #122, opened 2021-05-28
fix(accessibility): added aria labels to anchors that confuse screen …

PeculiarGoat:
…readers

https://dev.azure.com/S72/SHIFT72/_workitems/edit/4150

You could argue that these are false positives on SiteImprove because the anchors do in fact have content inside them but the screen readers don't handle them nicely.  If the anchor has an aria-label (`<a aria-label="The Matrix"><div>stuff</div></a>`) or some text before the other content (`<a>The Matrix<div>stuff</div></a>`), the screen reader will simply say "The Matrix, link".  If the anchor doesn't have either of those, then the screen reader will do a full info dump of the contents starting with availability labels, multiple instances of the film title, and repeating the word "link" at the end of every line _even if it is technically incorrect_.

In the interest of not introducing breaking changes for accessibility _just yet_, I think aria-labels are the quickest and safest solution.

### DIFF
--- a/site/templates/bundle/partial.jet
+++ b/site/templates/bundle/partial.jet
@@ -1,7 +1,7 @@
 {{import "../items/tagline.jet"}}
 
 <div class="partial partial-bundle">
-  <a href="{{routeToSlug(bundle.Slug)}}">
+  <a href="{{routeToSlug(bundle.Slug)}}" aria-label="{{bundle.Title}}">
     <div class="poster poster-{{config("default_image_type")}}">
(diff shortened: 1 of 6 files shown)


curtisadams28 (approved):

stajs:
I'm 70% sure that you shouldn't have a block level element inside the anchor and that is what is throwing the reader off - the invalid HTML of `<a><div/></a>`. This is probably a worthwhile bandaid in place of a more onerous restructuring of the page.

PR #124, opened 2021-05-31
Ab 4148 link color

smitsanghvi21:
https://dev.azure.com/S72/SHIFT72/_workitems/edit/4148/


PeculiarGoat (changes requested):
Couple of things...

site/styles/_meta-detail.scss
```diff
@@ -326,15 +326,17 @@
...
+      background-color: transparent;
       text-decoration: underline;
       animation: fadein 2s;
       font-size: 12px;
       -moz-animation: fadein 2s; /* Firefox */
       -webkit-animation: fadein 2s; /* Safari and Chrome */
       -o-animation: fadein 2s; /* Opera */
       @include hover-focus {
-        color: rgba($body-color, 0.8); 
+        color: $secondary;
```
PeculiarGoat, line 339: This should be `$body-color`... the client's secondary colour could easily be worse than the primary colour!

site/styles/main.scss
```diff
@@ -349,3 +349,24 @@ s72-classification-label {
 h1, h2, h3, h4, h5, h6, p, .title {
   word-wrap: break-word;
 }
+
+p, .s72-prompt, .s72-searchresults-empty-content, span, .s72-modal-body .s72-shopping-done {
```
PeculiarGoat, line 353: One per line please!  Putting loads of stuff on one line makes future diff-checking a pain in the butt!

site/styles/main.scss
```diff
@@ -349,3 +349,24 @@ s72-classification-label {
...
+a {
+  color: $body-color;
+  text-decoration: underline;
+  background-color: transparent;
+  font-weight: bold;
+  }
+  a:hover, a:focus {
```
PeculiarGoat, line 360: Please use the Sass ampersand e.g:
```
a {
  // styles...
  &:hover,
  &:focus {
    // styles...
  }
}
```

site/styles/main.scss
```diff
@@ -349,3 +349,24 @@ s72-classification-label {
...
+    color: $secondary;
```
PeculiarGoat, line 361: `$body-color`!

site/styles/main.scss
```diff
@@ -349,3 +349,24 @@ s72-classification-label {
...
+    text-decoration: none; }
+
+a:not([href]):not([tabindex]) {
```
PeculiarGoat, line 364: Not sure I understand what all this `not` stuff is for... carousel left/right buttons?

PeculiarGoat (changes requested):
It would be better if you made a link mixin and then include it in the other scss files where it is relevant.  We shouldn't be using `main.scss` to override styles that are defined in the other files.

site/styles/main.scss
smitsanghvi21: Shifted most of it in the relevant files. Some of them don't have a separate file, eg: comes under span or a div

kibble.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "core-template",
   "version": "0.0.1",
-  "siteUrl": "https://staging-abccinemas.screenplus.co",
+  "siteUrl": "https://abccinemas.screenplus.co",
```
PeculiarGoat, line 4: Assuming you committed this by accident? lol

site/styles/_mixins.scss
```diff
@@ -23,3 +23,17 @@
...
 }
+
+.a-links{
+    color: $body-color;
+    text-decoration: underline;
+    background-color: transparent;
+    font-weight: bold;
+}
+.a-links-hover-focus{
+    &:hover,
```
PeculiarGoat, line 34: These can be part of a-links... I don't see any point to them being a separate mixin.  Or am I missing something?

site/styles/_mixins.scss
```diff
@@ -23,3 +23,17 @@
     z-index: 8;
     @include gradient-x( rgba($left-gradient-bg-color, 1), transparent, 0%, 100%);
 }
+
+.a-links{
```
PeculiarGoat, line 27: Can you please use the [mixin syntax](https://sass-lang.com/documentation/at-rules/mixin) so there is no doubt what this is for!

PeculiarGoat: You will need @include in the places you use it instead of @extend too.

PeculiarGoat (changes requested):
Better... but I still have a few complaints lol

site/styles/_mixins.scss
smitsanghvi21: Initially I thought to use mixins and includes like the sass examples, but our default styles don't seem to be aligned with that. I mean instead we declare mixins in the _mixins.scss and use @extend to use it.

kibble.json
smitsanghvi21: Yea will change that

site/styles/_mixins.scss
smitsanghvi21: Some of the a links have a different hover/focus styles and merging the above ones into that would affect them. For ex, the nav a links have different hover combination

site/styles/_mixins.scss
PeculiarGoat: Yeah the existing stuff isn't great... but we should everything properly going forwards!

site/styles/_mixins.scss
PeculiarGoat: I think overriding the mixin styles for those elements is nicer than having to include 2 mixins on all the other anchor elements.

package-lock.json
```diff
@@ -1,8 +1,5821 @@
 {
   "name": "@shift72/core-template",
-  "version": "0.1.0-alpha.1",
-  "lockfileVersion": 1,
+  "version": "0.1.0-alpha.2",
```
PeculiarGoat, line 3: What's all this about? lol

site/styles/_shift72.scss
```diff
@@ -2,6 +2,7 @@
 // s72-slider styles in _slider.scss
 // shopping modal styles in _shopping.scss
 // user devices styles in _devices.scss
+@import '_mixins.scss';
```
PeculiarGoat, line 5: Does this import need to be here?  You didn't put it in the other files?

site/styles/main.scss
```diff
@@ -349,3 +349,9 @@ s72-classification-label {
 h1, h2, h3, h4, h5, h6, p, .title {
   word-wrap: break-word;
 }
+
+p,  span {
```
PeculiarGoat, line 353: Shouldn't be in this file!  This file is for overriding the core styles underneath (equivalent to local scss in shift72-template).

PeculiarGoat (changes requested):

site/styles/_shift72.scss
smitsanghvi21: Was getting an error something like mixin doesn't exist in this file and not in the other files. Importing it fixes the issue

package-lock.json
smitsanghvi21: Probably gets autoupdate when pulling changes or running npm i. It's the npm core package version

site/styles/main.scss
smitsanghvi21: `<a>` tag doesn't come directly under `p` and `span` in our scss files. Had to override this because sometimes the `span a {` or `p a {` gets autogenerated when a user adds something through admin or an input box.

PeculiarGoat: You can add `<a>` tags under `.meta-detail-synopsis`, `.page-header`, etc!

site/styles/_shift72.scss
PeculiarGoat: I think it might be because `_shift72` is imported before `_mixins` in `main.scss`... if you put the mixins import above the shift72 one, you should be able to remove the import from this file.

package-lock.json
PeculiarGoat: There's almost 6000 lines changed in this file... me and Curtis' PRs didn't have any changes to this file (or main.js.map).  What are you doing differently?

site/styles/main.scss
smitsanghvi21: It might still miss some things like if a link is added in the purchase modal or the bonus content section, etc. We would need to first identify each unique instance of <a> that would be used in that case.

package-lock.json
smitsanghvi21: Main.js.map would get rendered when there' any change. Can be just ignored or I can put it in gitignore. I think the last release is the cause of updates in the package-lock.json. It would work if you test locally.

site/styles/main.scss
PeculiarGoat: I think that's a better way.  Does @stajs have an opinion?

smitsanghvi21: Moreover if we just add <a> directly under `.meta-detail-synopsis, .page-header` it might also add the style if there are any buttons including the custom field btn, pricing, signup, etc

PeculiarGoat: So add more specific selectors!

site/styles/_mixins.scss
```diff
@@ -23,3 +23,15 @@
     z-index: 8;
     @include gradient-x( rgba($left-gradient-bg-color, 1), transparent, 0%, 100%);
 }
+
+@mixin a-links{
+    color: $body-color;
+    text-decoration: underline;
+    background-color: transparent;
+    font-weight: bold;
```
stajs, line 31: The underline is enough, don't need to be bold.

```suggestion
```

site/styles/_mixins.scss
```diff
@@ -23,3 +23,15 @@
     z-index: 8;
     @include gradient-x( rgba($left-gradient-bg-color, 1), transparent, 0%, 100%);
 }
+
+@mixin a-links{
+    color: $body-color;
```
stajs, line 28: I think we should have a `$link-color` that might initially be set to `$body-color` but gives us an easy override point if we need to have it be different.

site/styles/_search.scss
```diff
@@ -46,6 +46,9 @@
   .s72-searchresults-empty-content {
     font-size: 14px;
     font-weight: $font-weight-bold;
+    a{
```
stajs, line 49:
```suggestion
    a {
```

site/styles/_mixins.scss
```diff
@@ -23,3 +23,15 @@
     z-index: 8;
     @include gradient-x( rgba($left-gradient-bg-color, 1), transparent, 0%, 100%);
 }
+
+@mixin a-links{
```
stajs, line 27:
```suggestion
@mixin a {
```

site/styles/_mixins.scss
```diff
@@ -23,3 +23,15 @@
...
+    color: $body-color;
+    text-decoration: underline;
+    background-color: transparent;
+    font-weight: bold;
+    &:hover,
```
stajs, line 32:
```suggestion
       &:hover,
```

site/styles/_shift72.scss
```diff
@@ -226,4 +226,7 @@ s72-cant-be-watched {
 .s72-prompt {
   @extend p;
   @extend .text-center;
+  a{
```
stajs, line 229:
```suggestion
  a {
```

site/styles/_shopping.scss
```diff
@@ -190,6 +190,9 @@
       margin: 0 auto;
     }
   }
+  a{
```
stajs, line 193:
```suggestion
  a {
```

site/styles/main.scss
stajs: I think this could be a little dangerous leaving it here, e.g. for any links styled as buttons where you have a different hover effect than underlining the text.

stajs: For the case where the user has added a link in Admin for the synopsis, could we not have a more specific container to limit the selector?

stajs (changes requested):
I wasn't so keen on `a-links`, but struggle for alternative. I don't know if `a` is too short, but perhaps `a-behaviour`?

I think we should factor in an extension point for link colors.

site/styles/_mixins.scss
curtisadams28: FYI - We use these variables for font weights now:
`$font-weight-light: 300 !default;
$font-weight-bold: 500 !default;`

PeculiarGoat (changes requested):
Looking much better :+1: 

I think you need to add styles to target links in all of Admin-editable fields too though.  (film/tv/bundle synopsis, content/curated/collection text, etc).

site/styles/_variables.scss
```diff
@@ -97,6 +97,8 @@ $signup-border-color-hover: transparent !default;
 $signup-background: $button-background !default;
 $signup-background-hover: $button-background-hover !default;
 
+//Link color
```
stajs, line 100: I don't think the comment adds value.
```suggestion
```

stajs (commented):
Agree with @PeculiarGoat - do we have selectors we can target for links in user-editable text like synopsis etc.

smitsanghvi21:
@PeculiarGoat  @stajs Added selector to target synopsis (.meta-detail-synopsis - p - a) in the _meta_detail file. Would add it for other custom values too.

smitsanghvi21:
@stajs @PeculiarGoat added at few more places. Tested it on films, bundles, tv, custom page, collection, modal.

PR #131, opened 2021-06-02
Ab#4144 insufficient colour contrast

curtisadams28:


PeculiarGoat (changes requested):
There are more places with grey text on black (footer copyright statement, purchase modal text, etc).  Can you go through the site and try to find all of the places that need these changes please :)

Also, how did you decide on 0.8?

curtisadams28:
@PeculiarGoat I just went through what was recommended on the card. I chose 0.8 because they wanted 0.9 but there isn't a noticable diffirence between 0.9 and 1.0.

curtisadams28:
@PeculiarGoat According to the tool I'm using, the purchase modal etc have enough contrast. It's just the meta-item tags in the carousel and detail pages because they have image backgrounds. The card only mentions the meta-tags too.

PeculiarGoat:
@curtisadams28 Those other places may have enough contrast already... but we do want our < 1.0 opacity elements to be consistently opaque across the whole site.

curtisadams28:
@PeculiarGoat The latest commit should have changed all the opacity values for text to 0.8.

PeculiarGoat (approved):
Nice!

Better sort out that merge conflict!

PR #132, opened 2021-06-03
change font size to 12, add it in variable

smitsanghvi21:
https://dev.azure.com/S72/SHIFT72/_workitems/edit/4154/


PeculiarGoat (changes requested):
I know that the TIFF document doesn't show them, but there are more `font-size: 10px;` styles in core-template that should probably also be changed.  Feel free to run things past the design team if it ends up looking weird.

PeculiarGoat (approved):
Nice work!

Probably a good idea to add a list of things that have new font sizes for the ADO card to make Bryny's life easier.

PR #133, opened 2021-06-04
feat: added variables for search bar left position

PeculiarGoat:
The search bar's md left value was lowered when the new placeholder logos were added.  This means that 2x client logos overlap the search bar in md sized windows.  These new variables provide a quick and easy way to change the search bar position on client sites without needing to manually override the css.

As much as I dislike adding new variables, this seems like the easiest way to solve this problem while we get the new nav sorted.

### DIFF
--- a/site/styles/_nav.scss
+++ b/site/styles/_nav.scss
@@ -296,7 +296,7 @@
 // Header Form Search
 .navbar-nav-search {
   position: absolute;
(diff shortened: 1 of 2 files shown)


curtisadams28 (approved):
Looks good 👍

smitsanghvi21 (approved):

PR #135, opened 2021-06-10
fix(accessibility): added missing config check to bundle partial title

PeculiarGoat:
https://dev.azure.com/S72/SHIFT72/_workitems/edit/4153?src=WorkItemMention&src-action=artifact_link

'Fixes' the bundle title header that shows in search results.

TBH this fix is sidestepping the issue because it just stops the header from showing (I don't think any clients have this config on anyway).  But this brings the bundle partial in line with the other partials... SiteImprove couldn't see them with the config off.

We can come back to this later for a redesign!


l0ud0gg (approved):
looks legit

PR #137, opened 2021-06-16
fix: changed tagline from <ul> to <div> to fix hidden margin issue

PeculiarGoat:
https://dev.azure.com/S72/SHIFT72/_workitems/edit/3852

Decided to have another look at this task.  The mystery margin was caused by the first `<li>` containing an empty `<s72-classification-label>`... not exactly sure why it was doing that but I suspect something to do with the Bootstrap list item styles.  Fixed by changing from an `<ul>` to a `<div>` full of `<span>`.

Pro-tip: hide whitespace changes to make the jet diff easier to understand.

### DIFF
--- a/site/templates/items/tagline.jet
+++ b/site/templates/items/tagline.jet
@@ -1,63 +1,48 @@
-{{block metaItemTagline(genres=true, genresLimit=-1, classification=true)}}
-  <ul class="meta-item-tagline">
-
-    {*<li class="meta-item-tagline-item">
-      <span class="meta-item-classification meta-item-classification-mature">M</span>
-    </li>*}
-
-    {{if classification==true}}
-    <li class="meta-item-tagline-item meta-item-tagline-item-classification">
+{{ block metaItemTagline(genres=true, genresLimit=-1, classification=true) }}
+  <div class="meta-item-tagline">
+    {{ if classification==true }}
       <s72-classification-label data-slug="{{.Slug}}" data-layout="tooltip"></s72-classification-label>
-    </li>
-    {{end}}
+    {{ end }}
 
-    {{if genres == true}}
+    {{ if genres == true }}
       {{ if isset(.Genres) && len(.Genres) > 0 }}
-
         {{ itemGenres := .Genres }}
         {{ if genresLimit > -1  && len(itemGenres) > genresLimit }}
           {{ itemGenres = itemGenres[:genresLimit] }}
         {{ end }}
 
-        {{if isset(itemGenres)}}
-          <li class="meta-item-tagline-item">{{itemGenres}}</li>
-        {{end}}
+        {{ if isset(itemGenres) }}
+          <span class="meta-item-tagline-item">{{ itemGenres }}</span>
+        {{ end }}
 
-        {{if isset(.Runtime) || isset(.Episodes)}}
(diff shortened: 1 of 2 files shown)


smitsanghvi21:
@PeculiarGoat Seems to be properly aligned on [staging abc](https://staging-abccinemas.screenplus.co/film/the-happy-prince/) when I remove the classification.

PeculiarGoat:
@smitsanghvi21 look closer... it's not aligned!

site/templates/items/tagline.jet
```diff
@@ -1,63 +1,48 @@
...
-      <li class="meta-item-tagline-item">{{i18n("episode_count", len(.Episodes))}}</li>
+        {{ if isset(.Runtime) || isset(.Episodes) }}
+          <span class="meta-item-tagline-divider">•</span>
+        {{ end }}
+      {{ end }}
+    {{ end }}
+
+    {{ if isset(.Runtime) }}
+      {{ if .Runtime > 60 }}
+        <span class="meta-item-tagline-item">{{ .Runtime.Hours() }}{{ i18n("runtime_hours") }} {{ .Runtime.Minutes() }}{{ i18n("runtime_minutes") }}</span>
```
stajs, line 34: Is there a reason why the jet bits have been fluffed out with delimiter spaces? I assume an editor auto-format? It makes the PR diff hard to notice the actual changes.

I acknowledge [the official docs](https://github.com/CloudyKit/jet/blob/master/docs/syntax.md) 95% support the extra-fluffiness, but this line in particular I find it harder for my eye to spot where there should/shouldn't be a space in the rendered output and it also word wraps in a worse place. Perhaps we should take this to the team for a wider dev discussion and add an editorconfig to enforce a style.

stajs (changes requested):
Only requesting change to force a conversation on syntax.

site/templates/items/tagline.jet
PeculiarGoat: I do it because I find it harder to read without the spaces.

PeculiarGoat:
Approve approve approve!

PeculiarGoat:
Approve approve approve approve!

l0ud0gg (approved):
ran it up and tested against staging-abc. tried a few browsers and resolutions and looks good to me :)

smitsanghvi21 (approved):

stajs:
@l0ud0gg @smitsanghvi21 @curtisadams28 @brynyShift72 @PeculiarGoat Functionally this is fine, but I would like to see us agree on the syntax convention. This introduces a style of extra spaces in the jets, e.g. `{{ foo }}` vs what we have for 99% of existing usage with `{{foo}}`. The argument over what is more readable is subjective - @PeculiarGoat prefers spaces, I prefer sans-spaces. To me, it's akin to having `< div >`, a less consice syntax. I find it harder to see where the spaces should appear in the rendered output:

```
{{ .Runtime.Hours() }}{{ i18n("runtime_hours") }} {{ .Runtime.Minutes() }}{{ i18n("runtime_minutes") }}
```

smitsanghvi21:
I find `{{if foo}}` easier to read than `{{ if foo }}`. The other drawback  of spaces could be that it's easier to overlook and add spaces like this in the translation keys too and make a mistake.

stajs (approved):

PR #147, opened 2021-06-27
fix: hover preview text for content pages is now center aligned

PeculiarGoat:
https://dev.azure.com/S72/SHIFT72/_workitems/edit/4095

### DIFF
--- a/site/styles/_collections.scss
+++ b/site/styles/_collections.scss
@@ -162,6 +162,9 @@
               }
             }
           }
+          .meta-item-synopsis {
+            text-align: center;
+          }
           .synopsis {
             p {
               font-size: 12px;


imran-salim (approved):

site/styles/_collections.scss
```diff
@@ -162,6 +162,9 @@
               }
             }
           }
+          .meta-item-synopsis {
```
smitsanghvi21, line 165: Would be good to use the exact classname `meta-item-synopsis hidden-md`?

PeculiarGoat: It would be nice but I don't want to risk breaking changes for clients with custom nonsense for such a minor change.

smitsanghvi21 (approved):

stajs (approved):
I mean it does what it says on the tin, so all good. I took issue with the fact that it was a bug in the first place (as mentioned on the card), but happy for it to go in.

PR #151, opened 2021-07-07
feat(improvements): bundle page redesign

PeculiarGoat:
replaced bundle page HTML

replaced bundle page CSS

atomised bundle page jets to reduce impact of client customisation

### DIFF
--- a/site/styles/_meta-detail.scss
+++ b/site/styles/_meta-detail.scss
@@ -448,137 +448,149 @@
     padding-top: 80px;
   }
 }
-// Bundle Meta Detail
+
 .meta-detail-bundle {
-  // Meta Detail Background
   .meta-detail-bg {
     opacity: .4;
-    &:before {
+
+    &::before {
       display: none;
     }
-    &:after {
-      position: absolute;
-      width: 100%;
+
+    &::after {
+      @include gradient-y(rgba($body-bg, 0), $body-bg, 0%, 100%);
+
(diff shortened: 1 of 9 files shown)


PeculiarGoat:
A few notes:

- Some of the new jets in the `site/templates/bundle` directory will eventually be moved somewhere else to be shared with other pages.
- The CSS is built over what was already there rather than from scratch so it's a bit hacky.  I don't think it's worth starting over with the CSS until everything in core-template is has been atomised and all of the dud classes are stripped out of the html.

stajs (changes requested):
Requesting changes just cause I dislike the jet `{{ spacing }}`. I would prefer we collapse those.

Otherwise looks mostly good, although I question the value on the smaller atoms like title/synopsis.

PeculiarGoat:
The value of the smaller atoms is equal to all of the other atoms.  If the title/synopsis atoms are removed and a client requests a custom title/synopsis then we would have to freeze item.jet in time as is the current process.

PeculiarGoat:
@stajs Are you happy for this to be merged?  It will skip the need for that bundle button alignment task.

stajs (approved):

stajs:
@PeculiarGoat Yep, happy to merge.

I don't think all atoms are created equal. My dubiosity of their value is due to how unlikely I think a client will ask for changes of that.

PR #154, opened 2021-07-12
AB#4192 auto generate mobile support params / app links

l0ud0gg:
The existing problem is that the player relies on having an up to date supported-devices.json in order to show the user the correct information as to what they can and cant watch on / where to get the apps if they have them. At the moment its a manual process once apps are deployed but its a bit inconvenient having to do a site deploy to update it. 

I've added some logic into the supported-devices.json that will check;

1. IF mobile playback is enabled in uber and if so, add the support params for the mobile browsers.
2. IF the client has app links set in uber, and if so, adds the app param for the ios / android apps, otherwise leave as null. 

I've done some testing against ABC client 88 and set a couple of public configs (content  > configurations > app_link_ios  / app_link_android). I tried with and without the app links and mobile playback enabled / disabled / no config and seems to work fine. 

I suggest we add this into core template, and then remove from shift72-template so we dont override it. We will need to remove supported-devices.json from site local files as they are upgraded. not sure if there is a better way otherwise.

### DIFF
--- a/site/supported-devices.json.jet
+++ b/site/supported-devices.json.jet
@@ -1,7 +1,16 @@
-
 {
+  {{if config("player_allow_mobile_drm_content") == "true"}}
+  "support": {
+    "safari": {
+      "ios": "yes"
+    },
+    "chrome": {
+      "android": "yes"
+    }
+  },
+  {{end}}
   "apps": {
-    "android": null,
-    "ios": null
+    "android": {{if len(config("app_link_android")) > 0}}"{{config("app_link_android")}}"{{else}}null{{end}},
+    "ios": {{if len(config("app_link_ios")) > 0}}"{{config("app_link_ios")}}"{{else}}null{{end}}
   }
 }


l0ud0gg:
just an extra note to say the site needs to be rebuilt if configs change.

PeculiarGoat (approved):
Nice :+1:

stajs (approved):
N<i>iii</i>ce 👍

gscragg:
I would argue this functionality is better suited for being in the player. Then we don't need any client specific supported-devices.json files.

It already handles the [custom case of allowing mp4_fallback](https://github.com/indiereign/shift72-player/blob/master/app/scripts/players/players.service.js#L86).

I wasn't aware that the app links were being put in config, but if they are public, then it would be a simple task of getting them set via that method too.

l0ud0gg:
@gscragg yeah i was actually thinking that this morning also. The app links are not currently set in uber, this would be creating new config keys for that. I would much prefer that though even if it means adding all the keys into uber before its rolled out to the player and ignore the sites supported-devices files for automatic roll out. Theres quite a few sites that have inaccurate supported-devices.json files at the moment so this would ensure they are all correct in one sweep.

gscragg:
Right ok. So ignoring the app links, this change brings nothing to the table, outside of making sure clients don't have erroneous json files.

l0ud0gg:
i guess the benefit is that all config around what is supported is done via uber. it means the app team can set uber configs for the app links as they are rolled out and it doesnt require us to do site deploy (sometimes a big site upgrade) just to update the app links / remove support for mobile playback. Additionally, we are looking to bring back the app icons into the footer area of our sites so the same app links configs can be used to dictate if they should be shown and where they link to. in terms of the erroneous json files, yeah i think i have about 10 sites that need updating so i figured this would be a better path rather than needing to upgrade and deploy them all and avoid this in the future by making the process more smooth. Im not sure if there are any other reasons of having a local supported devices json if all the logic is in the player anyway.

gscragg:
Agree. If the app links are set in uber, they can be read via player and put in as needed.
The site templates can do the same for showing app links in the footer (or whereever).

stajs:
So what I'm hearing is we get player reading some new uber keys and we can retire the json file?

gscragg:
> So what I'm hearing is we get player reading some new uber keys and we can retire the json file?

Pretty much, I think there are a couple of older clients with custom files (ozflix) that might need investigating, but for the most part we shouldn't need them anymore.

l0ud0gg:
> I think there are a couple of older clients with custom files (ozflix) that might need investigating, but for the most part we shouldn't need them anymore.

Yeah cool. I'll have a dig around some of the older clients supported-devices files and check with Simon if he has any thoughts or concerns also. Otherwise maybe we could get jack to take a look at the player code to see if he can do a PR for these changes

smitsanghvi21 (approved):

imran-salim (approved):

l0ud0gg:
We've taken a different approach and adapted the player logic instead to generate the supported device matrix / app links via uber. This PR is now redundant. Further PR to follow to remove supported-device.json for sites going forward.

PR #155, opened 2021-07-12
Linters

PeculiarGoat:
- Added `stylelint` and `prettier` for scss linting.  The `stylelint-config-sass-guidelines` and `stylelint-config-prettier` configs give us a pretty good ruleset out of the box and basic issues can automatically be fixed.
- Added `eslint` for js linting on default settings.  Can also fix basic issues automatically.

Should be pretty simple to setup a drone yaml to run these linters when a PR is made/new commits are added to a PR and then we can start enforcing some consistency.

What do you think @stajs ?


stajs:
Looks good - can we get a team demo/discussion including drone?

gscragg:
Nice @PeculiarGoat.

On any new js/frontend shiz I have been using the configs from [here](https://github.com/shift72/cast-receiver) for lint, styling and editor setup. Also contains a custom vscode settings file for running eslint on save.

It uses googles eslint config which isn't for everyone, but I think it helps to enforce a minimum style on new code.

PeculiarGoat:
@stajs @gscragg

I've decided on `eslint-config-preact` we want Relish hacks/overrides in core-template sometimes (and that Google config has some weird stuff in it :P).  Also moved all the configuration into `package.json` so we don't have a bunch of random files.  I'm not super keen on linting-on-save with my potato machine but we can see what the rest of the team thinks.  `prettier` is now integrated properly so that can handle formatting before the linters handle style without overlapping rules.

Graham can you please add core-template to Drone?  I don't think I am special enough to do it anymore.

stajs:
I prefer separate files rather than putting more and more in `package.json`, however happy to go with the established pattern.

I also like lint on save - vscode has some settings heirarchy like Workspace Settings > User Setting > Default Settings, so we can possibly have it lint in user settings and if you want to opt out, override it in workspace settings.

stajs:
Hmm, nevermind - the settings override like I explained, but not how I was thinking of using them.

PeculiarGoat:
I think the extensions for these packages in VS Code can be setup to lint-on-save if anyone wants that.

PeculiarGoat:
@stajs @gscragg

Graham suggested trying Github Actions instead of Drone and it seems alright!  You should be able to see that "All checks have failed" because I haven't done the actual linting as part of this PR.

Are we happy for this to be merged now?

stajs (approved):
LGTM

PR #161, opened 2021-07-16
fix: added focus ring to language selector

PeculiarGoat:
https://dev.azure.com/S72/SHIFT72/_workitems/edit/4735

Looks like the new nav has a slightly different approach to focus rings than the old nav so the language selector styles needed a little tweak.

### DIFF
--- a/site/styles/_language-selector.scss
+++ b/site/styles/_language-selector.scss
@@ -30,6 +30,10 @@ s72-language-selector {
     &:focus {
       color: $primary;
     }
+    &:focus-visible {
+      outline: 1px dotted $body-color;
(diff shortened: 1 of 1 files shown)


curtisadams28:
@PeculiarGoat It still misses the language selector when I tab through. I think it's because `.s72-dropdown-toggle` is a `<div>` so it is skipped over. On the subnav dropdowns it is an `<a>`

PeculiarGoat:
@curtisadams28 did you test it with the latest Relish locally?  I've already sorted the <a>s.

curtisadams28 (approved):
I tested it with Relish. Looks good.

PR #162, opened 2021-07-18
App store badges

curtisadams28:


site/templates/footer/app-badges.jet
```diff
@@ -0,0 +1,13 @@
+{{ block appBadges() }}
```
stajs, line 1: We went through the team bloodletting for the holy whitespace war, I feel it would be remiss of me not to request striking down the unholy spaces of Satan.

stajs (approved):
... apart from the blasphemy.

PR #163, opened 2021-07-19
AB#4858 Added Sign Out Links

curtisadams28:
I added links inside the s72-sign-out component so that you can select it with keyboard controls.


PeculiarGoat (changes requested):
Couple of things:

- Please use `href="#"` so it is explicit that these links don't go anywhere (the Relish component overrides redirects here anyway).
- The focus styles for the signout button (for mobile and desktop) are different to the other account buttons... probably want to fix that.

PeculiarGoat:
There are still focus ring issue:

- In desktop view when tabbing through the account menu, the signout button's focus ring is a different shape/size to the others and the hover colour doesn't get activated.
- In mobile view, the same happens but maybe that signout button should have `display: none;` since there is the extra mobile signout button ... more duplication :(

stajs (approved):
I don't see anything wrong here - @PeculiarGoat do you still need your comments addressed?

curtisadams28:
@PeculiarGoat I've fixed the issues you mentioned. Not sure why that check is failing.

PeculiarGoat:
When I run it locally I get this:
image

Guessing that wasn't your intention lol.

Also not seeing any difference with the borked focus rings/colours on the dropdown signout button.

PeculiarGoat:
Also, don't worry about the checks... we are going to have to live with some failures while I figure out the linter kinks.

curtisadams28:
@PeculiarGoat The site must have been cached when I reordered the css. Fixed now.

PeculiarGoat (approved):
Nice work!  :+1:

PR #164, opened 2021-07-19
fix: error message background is now red

PeculiarGoat:
https://dev.azure.com/S72/SHIFT72/_workitems/edit/4465

Not sure how I feel about this one.  MaoriNow's problem isn't due to a bug on our end... it's because they are trying to use light _and_ dark themes on the same site and we don't support that.  Now we have an extra colour variable to maintain (and give us a headache on the next red background site) when the previous error colour worked fine and worked with any background colour.  :(


curtisadams28:
Having the background and text colour as variables is enough IMO. That way we can easily change it if we run into these edge cases.

PeculiarGoat:
I think that variables should only exist for common cases... edge cases should be solved with custom css (since they are literally customisations).

curtisadams28 (approved):
Looks good to me

site/styles/_forms.scss
```diff
@@ -51,7 +51,7 @@ s72-activatedevice-form,
 }
 // Error Messaging
 .s72-error-message {
-  background: rgba($body-color, 0.1);
+  background: $error-background;
   color: $body-color;
```
stajs, line 55:
```suggestion
  color: $error-color;
```

site/styles/_forms.scss
```diff
@@ -64,7 +64,7 @@ s72-activatedevice-form,
     height: 0;
     border-style: solid;
     border-width: 0 5px 5px 5px;
-    border-color: transparent transparent rgba($body-color, 0.1) transparent;
+    border-color: transparent transparent rgba($error-background, 1) transparent;
```
stajs, line 67: Don't like the faint border?

site/styles/_variables.scss
```diff
@@ -192,3 +192,5 @@ $font-family-base: 'Poppins', system-ui, -apple-system, "Segoe UI", Roboto, "Hel
 $font-weight-light: 300 !default;
 $font-weight-normal: 400 !default;
 $font-weight-bold: 500 !default;
+
+$error-background: #dc3545 !default;
```
stajs, line 196:
```suggestion
$error-color: $body-color !default;
$error-background: $body-bg !default;
```

stajs (changes requested):
The Maori Now bug was incidental to the discussion - the question being should errors be whatever $body-color might be (e.g. green), or should they be a UX best practice of red?

I think we need variables for both `color` and `background` to give us easy access to change if needed. If maintenance is a concern I don't mind them defaulting to `$body-color` and `$body-bg`, but that doesn't answer the orginal question.

stajs:
Let's see if linking works AB#4465

stajs (approved):

PR #165, opened 2021-07-20
body-bg-accent now changes depending on body-bg

curtisadams28:


site/styles/_variables.scss
```diff
@@ -37,11 +37,17 @@ $primary:         #35a0a8 !default;
 $secondary:       #4cd3dd !default;
 $body-color:      #fff !default;
 $body-bg:         #0f0f0f !default;
-$body-bg-light:   #1a1a1a !default;
+$body-bg-accent:   #1a1a1a !default;
 $hr-border-color: rgba($body-color, .5) !default;
 $list-group-bg: $body-bg !default;
 $list-group-border-color: rgba($body-color, .125) !default;
 
+@if lightness($body-bg) > 50% {
```
PeculiarGoat, line 45: This might be a bit nicer as a function that takes `$body-bg` as an argument so `$body-bg-accent` only has to be assigned once.

Any thoughts @stajs ?

PeculiarGoat (changes requested):

PeculiarGoat (approved):
Nice work!  :+1:

PR #166, opened 2021-07-25
fix: linter fixes for js and scss files

PeculiarGoat:
First pass of linter fixes.

Eslint is happy with the js files now.  Stylelint has a lot of complaints that can't be auto fixed for the scss files.  You see what the failures are by clicking through the checks menu below.

Some of these things look like they are going to be painful to fix... I think this might be a good opportunity to move to Dart Sass since node-sass is deprecated... IMO having the latest features/documentation will give us a better starting point to sort out these issues.


scripts/language-file-validator.js
```diff
@@ -16,20 +16,17 @@ const whitespacePaddedKeys = [
...
-];
-const emptyKeys = [
-  'classification_outro',
-  'plan_label_owned'
+  'classification_divider',
 ];
+const emptyKeys = ['classification_outro', 'plan_label_owned'];
 
-languageFilenames.forEach(language => {
+languageFilenames.forEach((language) => {
```
stajs, line 23: Is this linter mandated? I dislike the extra brackets for a single arg.

site/static/js/can-be-watched-button.component.js
```diff
@@ -11,32 +18,46 @@ export default class CanBeWatchedButton extends AppComponent {
...
-        return this.setState({ loaded: true });
-      }, utils.noop).then(() => {
-        let e = getComponentElement(this);
-        e.classList.toggle('s72-show', this.state.loaded && this.state.isOwned);
-      }, utils.noop);
+  updateState() {
+    if (
+      this.props.slug &&
+      this.app.shoppingService.isValidType(this.props.slug)
+    ) {
```
stajs, line 36: Are we wrapping because of a linter line length? I'm also not that keen on some of these wraps.

site/static/js/modernizr-custom.js
```diff
@@ -1,3 +1,86 @@
...
\ No newline at end of file
+!(function (e, n, s) {
```
stajs, line 3: Can we ignore auto  formatting this?

stajs (changes requested):
Couple of questions. I havent' reviewed all the style changes as there are way too many and first glance looks to be mostly alphabetizing.

stajs:
:100: move to Dart Sass, which is just `sass` [NPM module](https://www.npmjs.com/package/sass).

site/static/js/modernizr-custom.js
PeculiarGoat: :+1:

scripts/language-file-validator.js
PeculiarGoat: That's because of the prettier `arrowParens` option... I'll change that one for you.

site/static/js/can-be-watched-button.component.js
PeculiarGoat: This is caused by the prettier `printWidth` rule which is a bit awkward because [it doesn't work the way you might expect a linter line width rule to work](https://prettier.io/docs/en/options.html#print-width).  I've bumped it up to 100 which has fixed a lot of those overkill wrappings.

PeculiarGoat:
Moved to Dart Sass.  If you run the site you will now see a bunch of sass warnings fly by.  These are a warning for [this](https://sass-lang.com/documentation/breaking-changes/slash-div)... unfortunately I'm not sure if we can do anything about the ones originating from node_modules/bootstrap.  There is a `--quiet-deps` sass option which hides these warnings but it seems to have been removed from the sass executable :-1: .  We could also explore moving to Bootstrap 5 since that uses Dart Sass and I assume that means they wont be using deprecated features.

stajs (dismissed):

PeculiarGoat:
Checks are passing now.  I had to resort to adding linter disabling comments to some parts of the css because it's too messed up to be fixed just for this task... ideally we will replace the bad css at some point in the future (maybe when everything gets atomised).

Also added a few new rules to sidestep issues with Bootstrap as it looks like Bootstrap 5 has the same issues (so no point upgrading for this) and they have their own custom rule set that deviates from the sass guidelines rules.

The only thing that I haven't been able to suss out are the Dart Sass warning that show when you run the site... these are warnings for stuff that is happening in the Bootstrap node module and there doesn't seem to be a way of silencing them without silencing all warnings.

PeculiarGoat:
@stajs I've changed the way we handle local css in shift72-template (see [PR](https://github.com/shift72/shift72-template/pull/14)).

The idea is that:
- Core-template has a dummy local scss folder/file which is imported in main.scss.
- When shift72-template runs, it copies core-template scss into output.
- Then copies shift72-template local scss into output, overwriting the dummy local scss folder/file.
- Then runs Sass as normal.

site/styles/_buttons.scss
```diff
@@ -108,73 +106,70 @@
...
-        path {
-          fill: #000;
-          stroke: #000;
-        }
-      }
+    .s72-icon-play circle {
+      fill: $primary;
+      stroke: $primary;
+      stroke-width: 3px;
+    }
```
stajs: This is fine I guess, but was there a rationale against nesting? Because nesting seems more idomatic to me...

site/styles/_can-be-watched-button.scss
```diff
@@ -40,6 +40,7 @@
   }
 }
 
+/* stylelint-disable selector-max-compound-selectors */
```
stajs, line 43: I'm surprised we haven't needed this in more places TBH... I wonder if it would be better to bump the limit up to accomodate instead, I mean we know we've got a problem.

site/styles/_footer.scss
```diff
@@ -86,12 +86,13 @@ footer {
 
 .app-banner {
   align-items: center;
+  align-items: center;
+  border-bottom: 1px solid rgba($body-color, 0.2);
```
stajs, line 90: Duplication from a merge?

site/styles/_forms.scss
```diff
@@ -69,13 +69,13 @@ s72-activatedevice-form,
 
   &::before {
     border-color: transparent transparent rgba($body-color, 0.1) transparent;
+    border-color: transparent transparent $error-background transparent;
+    border-style: solid;
     border-style: solid;
     border-width: 0 5px 5px;
+    border-width: 0 5px 5px;
```
stajs: Duplication from a merge?

site/styles/_slider.scss
```diff
@@ -93,15 +85,12 @@ s72-slider.has-next.s72-slider-scroll::after {
...
 .slider {
-  // not sure why this is needed for sliders with only a couple of items, that dont require sliding.
   width: 100%;
 }
 
 .slider-fluid {
   position: relative;
-  // This changes the default value for s72-slider due to us using the fade-in/fade-out to show more items
-  // We instead use .slider-fluid as the relative positioned parent so the fades dont move when scrolling.
-  // removing this value will break those fades.
```
stajs, line 88: Aren't some of these comments still valuable?

stajs (commented):
Yikes, a `_legacy.scss`! Just got some questions before approving.

site/styles/_buttons.scss
PeculiarGoat: The rationale against **deep** nesting (_I'm not against sensible nesting_) is that it is less readable and less searchable, especially in long classes that require scrolling to read... some people view that as a [smell](https://csswizardry.com/2017/02/code-smells-in-css-revisited/#bem-mixes).  I think the 'BEM mixes' solution in that article is a good idea and we already do that a little bit in some places so perhaps we can gradually push in that direction more.

site/styles/_can-be-watched-button.scss
PeculiarGoat: I'd rather keep the limit where it is to force any new css to conform.

site/styles/_footer.scss
PeculiarGoat: Dammit!

site/styles/_forms.scss
PeculiarGoat: Dammit!

site/styles/_slider.scss
PeculiarGoat: Maybe :/

site/styles/_buttons.scss
stajs: Sure, but the refactored Sass is what I expect the original Sass to compile to in CSS anyway so we diverge from what I consider idiomatic Sass syntax and use more characters for the same result.

PeculiarGoat:
The `_legacy.scss` file just contains all the hack css that was in `main.scss` and shouldn't have been... couldn't think of a name for it!

site/styles/_buttons.scss
stajs: Note, I'm not attempting to defend the existing deep nesting.

PeculiarGoat: Isn't it fewer characters?  And fewer lines too?  :p

I did this refactor and a few others like it before I realised I could add those `stylelint-disable` comments... I can change it back and add the comment if you want but I reckon this way is 1% more human readable even if it's not the Sass way.

site/styles/_slider.scss
stajs: I mean, I'm not asserting they are, I don't know - just asking the question.

site/styles/_buttons.scss
stajs: More chars from having to duplicate `.s72-icon-play`. I don't see the advantage (or negative) of:

```
.s72-icon-play circle { }
.s72-icon-play path { }
```

... over:

```
.s72-icon-play {
  circle { }
  path { }
}
```

This is more words than needed over something so minor, leave it as is 👍

PR #167, opened 2021-07-26
latest kibble version, stripHTML method on tagline

matthewthomsonnz:
- This PR updates to the latest kibble which has a template method that strips html tags
- Here's [the card](https://dev.azure.com/S72/SHIFT72/_boards/board/t/Core%20team/Stories/?workitem=4407)
- Basically, clients have assumed they can add HTML to the 'Short description' field in admin, because they can do so in the synopsis below it.


stajs (approved):
LGTM 👍

PR #173, opened 2021-08-05
fix: prevent rollup.js from clearing console output when rebuilding

PeculiarGoat:
Might be a good idea to test this on your machines before approving to make sure it really works.


curtisadams28 (approved):
Works on my machine 👍

PR #177, opened 2021-08-11
fix: wishlist and trailer button margin/padding issues

PeculiarGoat:
https://dev.azure.com/S72/SHIFT72/_workitems/edit/5098

- removed padding from left of wishlist button so it aligns to left of container when trailer button is missing.
- added margin to right of trailer button so wishlist button isn't too close.
- removed margin and padding from wishlist button on posters so it aligns to the center.


site/styles/_buttons.scss
```diff
@@ -124,6 +124,14 @@
       fill: #000;
       stroke: #000;
     }
+  }
+  &.s72-btn-trailer {
```
stajs, line 128:
```suggestion
  }
  
  &.s72-btn-trailer {
```

site/styles/_buttons.scss
```diff
@@ -124,6 +124,14 @@
...
+    background: transparent;
+    border: 0 none;
```
stajs, line 130: Curious what the `none` is doing? I thought `0` would suppress it?

stajs (approved):
Minor Q.

imran-salim (approved):

site/styles/_buttons.scss
PeculiarGoat: That's been merged in from master... I didn't write it/don't know what it's for.

PR #184, opened 2021-08-30
Changed the light theme colours and added a dark overlay for the tablet menu

curtisadams28:


site/styles/_nav.scss
```diff
@@ -11,8 +10,16 @@
   top: 0;
   width: 100%;
   z-index: 9;
+  background: linear-gradient(rgba($top-gradient-color,1) 0%, rgba($top-gradient-color,0) ($navbar-brand-padding-y * 2) + 20px);
```
PeculiarGoat, line 13: I don't like this relying on `$navbar-brand-padding-y`... I think that variable may get removed when we add self-service branding.  Why not use a percentage?

site/templates/nav.jet
```diff
@@ -66,5 +66,6 @@
       </nav>
     {{end}}
   </div>
+  <div class="tablet-overlay"></div>
```
PeculiarGoat, line 69: I think a better class would be `s72-tablet-backdrop` to match our existing stuff.

PeculiarGoat (changes requested):
Couple of comments.  Also, you need to fix the linter complaints so the checks will pass!  `npm run css:lint`

site/styles/_nav.scss
curtisadams28: @PeculiarGoat This gradient goes down from the top of the `<header>` tag which I have made 100% the height of the screen when you open the menu. If the gradient had a percentage of 100%, then it would extend accross the screen. I chose not to hard code a value for the gradient because it wouldn't cover the nav properly if the logo is large.

curtisadams28: @PeculiarGoat I've fixed the issue by adding the gradient to `navbar-brand-container-tablet` instead of `<header>` when at the tablet breakpoints.

PeculiarGoat (approved):
Cool nice work!  :+1:

PR #185, opened 2021-08-30
update the cookie_banner_accept value in all language files

imran-salim:
[Bug 5133 Cookies Banner](https://dev.azure.com/S72/SHIFT72/_workitems/edit/5133)


smitsanghvi21:
@imran-salim I think we haven't come up with what needs to be done in this card. The design team wants some UI changes, not translation changes. Might be worth bringing this up in standup tomorrow.

imran-salim:
> @imran-salim I think we haven't come up with what needs to be done in this card. The design team wants some UI changes, not translation changes. Might be worth bringing this up in standup tomorrow.

Okay, I'll bring this up during standup tomorrow.

site/tr_TR.all.json
```diff
@@ -363,7 +363,7 @@
     "classification_outro": { "other": "" },
   
     "cookie_banner_message": { "other": "Bu site kullanıcı deneyimini artırmak için çerezler kullanmaktadır. Bu siteyi kullanırken, < çerez ve gizlilik ayarlarını kabul edin <a href=\"{{.CookiePolicyURL}}\">cookie usage</a>." },
-    "cookie_banner_accept": { "other": "Kabul Ediyorum" },
+    "cookie_banner_accept": { "other": "Kabul Etmek" },
```
PeculiarGoat, line 366: Doesn't this need to be one word to avoid the wrapping issue?

PeculiarGoat (changes requested):
FYI the ADO card has been changed... now only cares about preventing wrapping on the mobile button.  Not sure if there was much point changing the languages that are already one word.

site/tr_TR.all.json
stajs: I did have a thought/concern there could be a language where a single word wouldn't make sense, so we'd have to accomodate anyway. I'm not saying that is the case with this language (because how would we know!), just BTW.

PeculiarGoat (approved):
:+1:

PR #189, opened 2021-09-06
Ab#5163 vip signin button

PeculiarGoat:
https://dev.azure.com/S72/SHIFT72/_workitems/edit/5163

I'm not super keen on copy-pasting the signup-btn scss like this but it's probably the best way without redesigning all of the button scss :/


site/styles/_buttons.scss
```diff
@@ -40,6 +40,25 @@
...
+  white-space: initial;
+  @include media-breakpoint-up(xl) {
+    font-size: 16px;
+  }
+  @include hover-focus {
+    background-color: $signup-background-hover;
+    color: $signup-text-color-hover;
+  }
+}
+
```
stajs, line 61:
```suggestion
.btn.btn-signin-vip,
```

site/templates/nav/user_logged_out.jet
```diff
@@ -1,8 +1,10 @@
 {{block nav_user_logged_out()}}
   <s72-user-anon class="user-anon">
     <div class="vertical-line"></div>
-    <a href="{{ routeToPath("/signin.html") }}" class="btn btn-signin">{{i18n("nav_signin")}}</a>
-    {{if config("disable_user_signups") != "true"}}
+    {{if config("disable_user_signups") == "true"}}
+      <a href="{{ routeToPath("/signin.html") }}" class="btn btn-signin-vip">{{i18n("nav_signin")}}</a>
+    {{else}}
+      <a href="{{ routeToPath("/signin.html") }}" class="btn btn-signin">{{i18n("nav_signin")}}</a>
```
stajs, line 7: Looking at the config called `disable_user_signups` - I think it's possible that we would want this for other reasons than VIP sites, so I think we should not use `-vip` in the class if we can come up with something more generic. Perhaps `-cta` (for "call to action") or `-primary`?

Can we have the `if` inside the `class` or does the double-quotes play silly buggers? Thinking it might be nicer/concise rather than duplicating the tag.

stajs (changes requested):

site/templates/nav/user_logged_out.jet
PeculiarGoat: Okay I'll add `-primary`.

It's going to be a case of duplicating the tag or duplicating the `if` if you want that inside the the `class`.  IMO, duplicating the tag is simpler logic and easier on the eyes!

site/styles/_buttons.scss
PeculiarGoat: lolwut... that was a brain failure :/

stajs (approved):
Approving based on `-vip` becoming `-primary`.

PR #191, opened 2021-09-09
Added nav colour variables

curtisadams28:
A few things changed with this one:

- I added a number of colour variables so different parts of the nav can be changed independantly.
- `get-accent-color()` now uses `mix()` instead of `darken()`/`lighten()`. This better approximates having a colour like `rgba(255, 255, 255, 0.04)` over `body-bg`.
- I removed some variables that aren't being used anymore. Might be a good idea to go through and find them all at some point.


PeculiarGoat:
:+1:  looks good to me!

PeculiarGoat (approved):

PR #193, opened 2021-09-14
Ab#5507 no carousel when empty

l0ud0gg:


site/templates/collection/carousel.jet
```diff
@@ -23,5 +23,8 @@
 		<a class="right s72-carousel-control"><span class="fa fa-angle-right"></span></a>
 	</section>
 </s72-carousel>
+{{else}}
+<div class="no-carousel-padding">
```
l0ud0gg, line 27: Smit suggested we should indent this. The carousel component isnt indented in the upper section so i left as is for consistency. should they both be indented ideally inside the if / else statements?

PeculiarGoat: Yeah they should both be indented!

PeculiarGoat:
I'm not sure how I feel about this solution... I'm generally not keen on this sort of placeholder/spacer/do-nothing element.  It's a quick win, but I wonder if instead it's worth looking at how to make the homepage behave the same way as the other pages rather than giving it it's own solution like this.

Thoughts @stajs ?  Perhaps this solution can be temporary while the carousel redesign gets sorted?

stajs:
@PeculiarGoat I don't mind a band-aid while we wait for the paddings to be fixed.

PeculiarGoat:
@l0ud0gg I've figured out another solution and put it in my new carousel branch.  But you can totally add it here too if you want.

Add the following in `homepage.jet`:
Screenshot_2021-09-14_21-42-11

Add `no-carousel-hompage` to this rule in `_pages.scss`:
Screenshot_2021-09-14_21-42-57

No spacer elements required!

stajs:
Extending your idea, could always have a class (doesn't matter if it isn't used):

```html
<main id="main" class="page page-{{page.PageType|lower}} carousel-len-{{len(page.PageCollections[0])}}">
```

... and then target `.carousel-len-0` for padding. You won't need the extra Go code to declare a variable and `if` check.

---

But my preference would be to have a wrapper `<div>` in `carousel.jet` (reduced for brevity):

```html
<div class="carousel">{{if len(.Items) > 0}}<s72-carousel />{{end}}</div>
```

... and then in CSS:

```css
.carousel:empty { /*padding*/ }
```

Unfortunately current browsers don't support the latest spec to allow whitespace to match `:empty`, so it's an ugly butting up of the `{{if}}` and `{{end}}` inside the `<div>`. Otherwise, I'd combine the technique:

```html
<div class="carousel-len-{{len(.Items)}}">
  {{if len(.Items) > 0}}
    <s72-carousel />
  {{end}}
</div>
```

l0ud0gg:
sounds good. i'll close this PR as it can be done as part of the new carousel work instead . nice work

PR #198, opened 2021-09-21
Primary color fixes

curtisadams28:
ADO: https://dev.azure.com/S72/SHIFT72/_boards/board/t/Delivery%20team/Stories/?workitem=5270
Test Site: https://staging-test.shift72.com/

The task is to remove the primary colour from the hover state of a number of elements including:

- Subnav
- Trailer button
- Dropdowns

The changes make the site look worse IMO. The things we do for ~~accessibility~~ fixing clients poor decisions.


PeculiarGoat:
I'm pretty sceptical about this.  If the problem is that hovers are _sometimes_ harder to see, I don't see how making sure that hovers are _always_ harder to see is a good solution.  The whole point of the hover state is to make something stand out... not dim it.  Maybe we should have had a team discussion about this before the work was done?

curtisadams28:
@PeculiarGoat Good point. The more I worked on it, the less I liked the idea but I got some serious sunk cost fallacy.

curtisadams28:
I think I'll just start posting in slack whenever I start work on a design task.

PR #201, opened 2021-09-28
New meta-item-tagline jet + homepage padding fix when carousel is empty

PeculiarGoat:
- Homepage top-padding fix for when carousel is empty.
- `meta-item-tagline.jet` total rewrite:
  - Added release year to `metaItemTagline` and removed it from everywhere else (various titles).
  - Added more intelligent way of deciding how many divider dots to use (the old way was pretty easy to break and a total mess to read).
  - Added tidier way of showing info from the meta item.


site/templates/items/tagline.jet
```diff
@@ -1,48 +1,90 @@
...
+      {{if isset(.Genres)}}
+        {{itemGenres = .Genres}}
+      {{else if isset(.ShowInfo) && isset(.ShowInfo.Genres)}}
+        {{itemGenres = .ShowInfo.Genres}}
       {{end}}
+    {{end}}
 
-      {{if isset(.ShowInfo) && isset(.ShowInfo.Genres)}}
-        <span class="meta-item-tagline-item">{{.ShowInfo.Genres}}</span>
+    {*{ FIGURE OUT HOW MANY DIVIDERS }*}
```
stajs, line 12: Can we have non-shouty case plz

stajs (approved):
The logical structure is better, but would be curious if we can have a less verbose solution with an array of tag line items that are then joined by a divider (if the jet syntax could do it).

PeculiarGoat:
@stajs Unfortunately we have no way of converting the ints (release year, runtime, etc) into strings to put into a typed array in the jets.

PR #220, opened 2021-10-19
Choose the right width for md/lg logos

stajs:
Use the right logo size variable in the right breakpoint


stajs:
Good call, forgot about changelog

PR #222, opened 2021-10-21
plans/passes page styling

matthewthomsonnz:


site/en_AU.all.json
```diff
@@ -524,7 +524,7 @@
     "one": "Automatically renews every {{.Interval}}",
     "other": "Automatically renews every {{.Count}} {{.Interval}}s"
   },
-  "plans_page_header": { "other": "Available Plans" },
+  "plans_page_header": { "other": "Passes" },
```
additiverich, line 527: I think we should say "Plans" or "Available Plans" here and let them override it using a feature toggle or lang file

site/styles/_variables.scss
```diff
@@ -32,11 +32,29 @@ $container-max-widths: (
...
+  '5': ($spacer * 5),      // 50px
+);
+
 // Body / Core
 // ------------------------------------
 
-$primary: #35a0a8 !default;
-$secondary: #4cd3dd !default;
+$primary: #4d97ff !default;
+$secondary: lighten(#4d97ff, 7) !default;
```
additiverich, line 57: I know design want to mess with our default primary colour but I wonder if this is the time and/or place for that. I'll tag in @stajs and get his perspective.

site/styles/_buttons.scss
```diff
@@ -230,3 +230,16 @@
     margin-right: 5px;
   }
 }
+
+.btn-block .s72-btn {
+  @extend .btn-block;
+}
+
+.btn-light {
+  background-color: $gray-300;
```
stajs, line 239: I think these grays could have problems on some clients body colors - they probably need to be configurable.

EDIT: I don't know where you are using them so you might have it on some other element/background color - I haven't seen any designs.

site/styles/_variables.scss
```diff
@@ -32,11 +32,29 @@ $container-max-widths: (
...
+  'p5': ($spacer * 0.5),   // 5px
+  '1': $spacer,            // 10px
+  '1p5': ($spacer * 1.5),  // 15px
+  '2': ($spacer * 2),      // 20px
+  '2p5': ($spacer * 2.5),  // 25px
+  '3': ($spacer * 3),      // 30px
+  '3p5': ($spacer * 3.5),  // 35px
+  '4': ($spacer * 4),      // 40px
+  '4p5': ($spacer * 4.5),  // 45px
+  '5': ($spacer * 5),      // 50px
```
stajs, line 50: You using all of these? Or they just for completeness?

site/styles/_icons.scss
```diff
@@ -31,3 +31,7 @@
   @extend .fa;
   @extend .fa-search;
 }
+
+main svg {
+  fill: $primary;
```
stajs, line 36: This might have unintended consequences - we probably don't want to default every SVG in `main` to be primary.

stajs (changes requested):
Just concerned about every SVG being given a fill color.

matthewthomsonnz:
> Looks great on the page; can we check how it looks using the light template?
> 
> I've also flagged up potentially changing primary/secondary colours as part of this work.

Seems good
Screenshot from 2021-11-01 11-54-22

site/en_AU.all.json
matthewthomsonnz: - Reverted to Available Plans

site/styles/_variables.scss
matthewthomsonnz: - Reverted to original/master colours

matthewthomsonnz:
> Just concerned about every SVG being given a fill color.

Have used `.alert svg` as I think .alert is the closest bootstrap description for this element

And icons inside all alerts would be the same (else use specificity)

site/styles/_variables.scss
matthewthomsonnz: - Not all of them - have removed unnecessary ones

site/styles/_buttons.scss
matthewthomsonnz: Bootstrap makes this configurable? e.g. change to $gray-100,  $gray-900 etc per client. Or change .btn-light to use a custom bg color.

This shade of gray works for at least stark white/black.

stajs: By configurable I meant defining a custom variable for it, e.g. `$button-light-bg-color: $gray-300`. Otherwise the only way to change it is override with custom CSS e.g. `.btn-light { background-color: #ff0033; }`.

site/styles/_icons.scss
stajs: Not sure, but increases over time.

stajs:
I think if you resolve the conversations, this should be good to approve.

matthewthomsonnz:
> I think if you resolve the conversations, this should be good to approve.

Thanks @stajs have done. I added a bit more css though, if you could cast your eye over again sorry. Want to get your blessing on these two in particular as they're global:
1) [`<small>`  elem font size changed from 80% to 85% ](https://github.com/shift72/core-template/pull/222/files#diff-c7a109853705370c20be82551ad7c0b1ca7fec0ee478c206ae202feb3473e556R1-R3) to match designs
Screenshot from 2021-11-09 13-13-54 
Screenshot from 2021-11-09 13-14-45
2) [Every anchor that doesn't have a class gets an underline](https://github.com/shift72/core-template/pull/222/files#diff-c7a109853705370c20be82551ad7c0b1ca7fec0ee478c206ae202feb3473e556R5-R7) (very opinionated but will scope it if you would rather not)
Screenshot from 2021-11-09 13-13-25

site/plans.html.jet
```diff
@@ -18,35 +11,57 @@
...
-                <div class="plan-trial-period">{{i18n("shopping_info_trial_period_offer", p.TrialPeriodDays)}}</div>
-              {{ end }}
-            </div>
-          {{ end }}
-          <div class="plan-buttons">
-            {{ if p.Page }}
-              <button class="plan-page-link"><a href="{{ routeToSlug(p.Page.Slug) }}">{{i18n("plan_read_more")}}</a></button>
+        <div class="plan col-xs-6 col-lg-4 mb-3">
+          <div class="card h-100">
+            {{ if isset(p.LandscapeImage) && p.LandscapeImage != ""}}
```
stajs, line 19: A minor nit: the majority of the templates have no whitespace after/before the mustaches.

site/plans.html.jet
```diff
@@ -18,35 +11,57 @@
...
+                    <s72-plan-label data-slug="{{p.Slug}}" data-title="{{p.Name}}"></s72-plan-label>
+                  {{ end }}
+
+                  <h3 class="plan-name">{{p.Name}}</h3>
+
+                  {{ if p.HasExpiryDate() }}
+                    <p class="alert p-0 border-0">
+                      <svg xmlns="http://www.w3.org/2000/svg" width="16" height="16" fill="currentColor" class="bi bi-exclamation-circle" viewBox="0 0 16 16"> <path d="M8 15A7 7 0 1 1 8 1a7 7 0 0 1 0 14zm0 1A8 8 0 1 0 8 0a8 8 0 0 0 0 16z"/> <path d="M7.002 11a1 1 0 1 1 2 0 1 1 0 0 1-2 0zM7.1 4.995a.905.905 0 1 1 1.8 0l-.35 3.507a.552.552 0 0 1-1.1 0L7.1 4.995z"/> </svg>
+                      <small>
+                        {{ i18n("plan_expiry_date") }} <s72-time datetime="{{ p.ExpiryDate.UnixMilli() }}" />
```
stajs, line 44: FYI, the humanized date based off the browsers language is a current source of headaches with a large number of clients  not happy with the lack of translatableness. Just noting here as it's something we want to change for availability labels so this might also need a change later too.

stajs (approved):
Aside from the lolwut of `<small>` becoming larger, just a minor nit.

site/plans.html.jet
matthewthomsonnz: - removed space before/after moustaches

site/plans.html.jet
matthewthomsonnz: Cool. Not sure if you're aware - BAs are looking into a (manual human) translate service for these files, so that might eliminate some headache in the future.

PR #227, opened 2021-10-29
too many requests i18n support

matthewthomsonnz:
Adds support for too many requests error from recently added rate limiting on `/account_exists `


stajs (approved):
Approving on faith

PR #230, opened 2021-11-03
removed "Credit" from card

smitsanghvi21:
I think it shouldn't be having "Credit" card everywhere bc a user might think that they can only rent/buy an item if they have a credit card. In some countries like India there are different types of card like Gift card, Amazon pay card, Mobile wallet cards which are like a debit card and still allow online purchases. Also came up in this card: https://dev.azure.com/S72/SHIFT72/_workitems/edit/5937/

### DIFF
--- a/site/en_AU.all.json
+++ b/site/en_AU.all.json
@@ -249,7 +249,7 @@
   },
   "shopping_info_credit_bonus_title": { "other": "{{.BonusPercentage}}% Bonus" },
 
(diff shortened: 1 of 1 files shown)


stajs:
I agree but the wording approval needs to be run past @nevgarven. Also, need to update all other languages.

smitsanghvi21:
@stajs Should I close this PR then?

stajs:
No, just go through the right channels - I think it's a worthy change.

PR #237, opened 2021-11-23
AB#5536 Plans pages

additiverich:
As per [design feedback](https://dev.azure.com/S72/SHIFT72/_workitems/edit/5536)

- New translations for `availability_not_available` for Catalan, AU, Hungarian, Portuguese + BR, and Russian
- Free plans won't have pricing buttons (because they're generally assigned to users on sign up) so replace this with a Sign Up CTA for anonymous users
  - This CTA button uses the translation for `plan_free_link_text` for all languages

All passes page:
- 2rem border to h1 header
- get rid of border around cards
- wrap content of plans cards in a container to apply styles with (so the white areas are not all taking up the same height) and give these containers a background image and border radius
- paragraph spacing inside cards is different
- support for `s72-before` and `s72-after` tags on showcase page and passes page

Showcase page:
- hide separator in footer
- CTA field above footer should be #191919 rather than #262626.
- `s72-before` and `s72-after` tag for expiration date (inc. for free sites)
- hide the titles for all the collections

Not touching these:
- font sizes
- background images
- width of text elements
- layouts of titles in lists and sliders


site/plans.html.jet
```diff
@@ -59,9 +66,17 @@
...
                   {{ end }}
-                  <s72-pricing-buttons class="btn-block" data-slug="{{p.Slug}}" data-title="{{p.Name}}" />
+                  {{ if p.PlanType != "free" }}
+                    <s72-pricing-buttons class="btn-block" data-slug="{{p.Slug}}" data-title="{{p.Name}}" />
+                  {{ end }}
+                  {{ if p.PlanType == "free" }}
+                    <s72-user-anon>
+                      <a class="btn btn-block s72-btn-purchase" href="{{i18n("plan_free_link_url")}}">{{ i18n("plan_free_link_text") }}</a>
+                    </s72-user-anon>
+                  {{ end }}
```
matthewthomsonnz, line 76: `{{ if p.PlanType == "free" }} ...  {{ else }} ... {{ end }}`?

matthewthomsonnz:
Looks good, I'll let one of the deliveries approve though

additiverich:
Looping in my esteemed colleagues in the delivery team to weigh-in

site/plans.html.jet
```diff
@@ -59,9 +66,16 @@
                   {{ if p.Page }}
                     <a class="plan-page-link btn btn-light text-dark btn-block" href="{{ routeToSlug(p.Page.Slug) }}">{{i18n("plan_read_more")}}</a>
                   {{ end }}
-                  <s72-pricing-buttons class="btn-block" data-slug="{{p.Slug}}" data-title="{{p.Name}}" />
+                  {{ if p.PlanType == "free" }}
```
ashesnz, line 69: Maybe discuss while I was away, but are we allowing users to sign up even if the plan has already expired? Im guessing these buttons display after expiry date?

additiverich: We could wrap this in a <s72-before> tag so the CTA won't appear after the expiry date.

additiverich: @ashesnz Done

site/plans.html.jet
```diff
@@ -59,9 +66,24 @@
...
+                    {{ if p.HasExpiryDate() }}
```
stajs, line 70: I hesitate to suggest this, but just with the repeated blocks of `<s72-user-anon><a></s72-user-anon>` you _could_ potentially not check `p.HasExpiryDate()` and use a date variable that is either the plan expiry or a future date and then you could always use `<s72-before>`.

E.g. pseudo-code:

```html
{{expiry := p.HasExpiryDate() ? p.ExpiryDate : [tomorrow, next week, max date - dealers choice]}}
{{if p.PlanType == "free"}}
  <s72-before datetime="{{expiry}}">
    <s72-user-anon>
      <a>
    </s72-user-anon>
  </s72-before>
{{ else }}
  <s72-pricing-buttons>
{{ end }}
```

site/templates/page/curated.jet
```diff
@@ -15,7 +15,7 @@
   {{ plan = page.Plans[0] }}
 {{ end }}
 
-  <main id="main" class="page curated-page {{if len(page.PageCollections) > 0}}curated-collection-page{{end}}">
+  <main id="main" class="page curated-page {{if len(page.PageCollections) > 0}}curated-collection-page{{end}} {{if plan != 0}}curated-plan-page{{end}}">
```
stajs, line 18: Nit: I know you are following the existing style here, but can we change both to have the leading space in the `if` so that we don't end up with extra spaces when not used.

```suggestion
  <main id="main" class="page curated-page{{if len(page.PageCollections) > 0}} curated-collection-page{{end}}{{if plan != 0}} curated-plan-page{{end}}">
```

stajs (changes requested):
Apart from the `siteUrl`, just a bunch of kutus.

And to FYI one more, we have a mix of mustache whitespacing but we are conventioning on sans-leading/trailing whitespace, so prefer `{{blah}}` over `{{ blah }}`.

site/templates/page/page-content.jet
```diff
@@ -4,8 +4,22 @@
       {{text | raw}}
 
       {{if plan != 0 }}
-      <div>
+      <div class="plan-buttons">
+        {{ if plan.PlanType == "free" }}
+          {{ if plan.HasExpiryDate() }}
+          <s72-before datetime="{{ plan.ExpiryDate.UnixMilli() }}">
+            <s72-user-anon>
+              <a class="btn btn-block s72-btn-purchase signup-link" href="{{i18n("plan_free_link_url")}}">{{ i18n("plan_free_link_text") }}</a>
```
PeculiarGoat, line 12: Given that `plan_free_link_url` isn't actually a translation (and using it without `routeToPath` will always send you to the default language page and trigger a redirect), I would prefer the path hardcoded into the template `href="{{routeToPath("/signup.html")}}"`.  If it is supposed to be configurable, it would make more sense as a config that as a translation hack.

PeculiarGoat (changes requested):

site/plans.html.jet
additiverich: Might leave this as is; a line like

```
{{ expiry := plan.HasExpiryDate() ? plan.ExpiryDate.UnixMilli() : 4100670000000 }}
```

is borderline hackery even for me

additiverich:
Thanks @stajs @PeculiarGoat; your counsel appreciated as ever

PeculiarGoat (approved):
I'm happy.  Is @stajs ?

site/plans.html.jet
stajs: Well, when you put it like _that_ sure lol

I was thinking more `time.Now().AddDate(1, 0, 0)` or similar.

stajs (approved):
:shipit:

PR #239, opened 2021-11-30
AB#6089 award nominations on the front end

additiverich:
[Card here](https://dev.azure.com/S72/SHIFT72/_boards/board/t/Core%20team/Stories/?workitem=6089)
Designs [here](https://zpl.io/a3Jm18y)

- Changes to display an item's award nominations to complete Phase 1
- Not voting or results yet (Phases 2 & 3)

Homepage
 - carousel displays "in competition" if the title has at least one award nomination
 - meta item thumbnails display (solid) star icon if title has at least one nomination

Film page
 - List of nominations displayed between play / purchase buttons & availability label, and synopsis


site/templates/common/awards.jet
```diff
@@ -0,0 +1,31 @@
...
+      <div class="meta-award-nomination mb-2p5 col-12 col-sm-8 col-lg-6 col-xl-4">
+        <span><i class="meta-award-nomination-icon"></i></span>
+        <span class="font-weight-bold">{{ i18n("awards_nominated_for") }} {{ .DisplayLabel }}</span>
+      </div>
+    {{ end }}
+    </div>
+  {{ end }}
+{{end}}
+
+{{block inCompetition()}}
```
PeculiarGoat, line 14: I think one block per file is better.  Having several blocks here means that when I write an override for `awardNominations`, I also need to copy overrides for the other two blocks.

PeculiarGoat (changes requested):
One thing so far!

PeculiarGoat (changes requested):
All looks good except for one issue that I've found:

The star in the top-right corner of a poster clashes with the availability-status component if it has a long string inside.
image
This example is a real string from the Lithuanian translation file (I had to wiggle the screen size to get it to overlap though).

I'm not invited to the Zeplin workspace so can't see if this was considered by design... but I feel like we should do something about it!

additiverich:
@PeculiarGoat @matthewthomsonnz This might need a careful once over - I've wrapped the availability status (e.g. "SOLD OUT", "EXPIRED" etc etc) and the icon in a shared wrapper, and applied some flexbox styles to them to get them not to overlap and give them some breathing room. I've tried to be pretty specific with the CSS selectors to avoid any unintended consequences.

PeculiarGoat (approved):
Not sure how I feel about that text-overflow on the availability-status but it's not a show-stopper.  Everything else looks good to me!

PR #240, opened 2021-12-02
move genre after duration in tagline

matthewthomsonnz:
As requested by design - change order of tagline info so it goes duration -> genre -> year

Card [here](https://dev.azure.com/S72/SHIFT72/_boards/board/t/Core%20team/Stories/?workitem=4976)

### DIFF
--- a/site/templates/items/tagline.jet
+++ b/site/templates/items/tagline.jet
@@ -33,17 +33,6 @@
       <s72-classification-label data-slug="{{.Slug}}" data-layout="tooltip"></s72-classification-label>
     {{end}}
 
-    {{if genres && len(itemGenres) > 0}}
-      {{if genresLimit > -1 && len(itemGenres) > genresLimit}}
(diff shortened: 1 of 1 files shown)


stajs (approved):
It does what it says on the tin, but I can't help feeling this is micromanaging the design and we end up customizing the order for some clients so it won't always be this order.

stajs:
@matthewthomsonnz BTW, please request a review from the delivery team in future, unless you have need of specific eyes, it means we can share the PR load on the team.

matthewthomsonnz:
> It does what it says on the tin, but I can't help feeling this is micromanaging the design and we end up customizing the order for some clients so it won't always be this order.
> 
> EDIT: to expand, we have enough support requests to change this order that we have talked about making it self-servicable.

I'll mention this one also, I feel like there may be a (consultation?) step missing in our process somewhere around task creation.

PR #243, opened 2021-12-08
Custom font template

stajs:
Film at Lincoln Center (as well as lots of other clients) have a custom font they want to use. Previously this has required a custom `application.jet`, but this PR adds a `font.jet` that is now included in `application.jet` meaning a smaller customisation is required to change the font used. At some point this could even be a self-service change.

The PR has a mass of changes because I also took the liberty to re-organize a bunch of templates sitting in `/site/templates` -> `/site/templates/application` including `application.jet` and header/footer/nav templates. This touches a lot of files but it's all simple path changes.

I dont know why `main.css.map` has changed since I haven't touched the styles. I think we oughta ignore that file but leaving for discussion.


PeculiarGoat (approved):
No Google doc?  No elab?
evil-laughter

PR #250, opened 2021-12-15
Sponsor banner

curtisadams28:
I have added the sponsor banner to core-template behind the config `show_sponsor_banner` in User > Configurations. I'm leaving things simple for now as  I copy pasted Jacks code almost verbatim but we should consider revisiting this with self-service in the future.


stajs:
Need a description and changelog plz

kibble.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "core-template",
   "version": "0.0.1",
-  "siteUrl": "https://abccinemas.screenplus.co",
+  "siteUrl": "https://staging-test.shift72.com",
```
additiverich, line 4: Are we moving to use this site rather than abccinemas

curtisadams28: I left that in by mistake. I'll fix it now.

stajs (approved):
Approving, but would appreciate removing the mustache whitespace, e.g. prefer `{{end}}` over `{{ end }}`.

curtisadams28:
This has been put on hold until the new [image model](https://dev.azure.com/S72/SHIFT72/_workitems/edit/5093) is implimented.

curtisadams28:
I have asked around and it doesn't look like the new image model will cause any issues with this.

PR #254, opened 2022-01-17
Added captions translation + split CC and Subs

brynyShift72:
#AB3859 - Captions appearing as subtitles: updated item.jet to split out captions and subtitles and display them separately, 
updated all translations files to use new "meta_detail_captions" translation key, place holders are subtitle string + [CC] as that symbol seems universal eg: fr-FR `"meta_detail_subtitles": { "other": "Sous-titres" },
  "meta_detail_captions": { "other": "Sous-titres [CC]" `

### DIFF
--- a/site/templates/film/item.jet
+++ b/site/templates/film/item.jet
@@ -91,15 +91,20 @@
               <p>{{ film.Languages }}</p>
             </div>
           {{end}}
-          {{if len(film.GetSubtitles()) > 0 }}
+          {{allSubtitles := film.GetSubtitles()}}
+          {{if len(allSubtitles) > 0 }}
             {{subtitles := ""}}
             {{captions := ""}}
-            {{range subTrack := film.SubtitleTracks}}
-              {{if len(subTrack) >= 3}}
-                {{if subTrack[2] == "caption"}}
-                    {{captions = captions ? captions + ", " + subTrack[1] : subTrack[1]}}
-                {{else}}
-                    {{subtitles = subtitles ? subtitles + ", " + subTrack[1] : subTrack[1]}}
+            {{if len(film.SubtitleTracks) == 0}}
+              {{subtitles = allSubtitles}}
+            {{else}}
+              {{range subTrack := film.SubtitleTracks}}
+                {{if len(subTrack) >= 3}}
+                  {{if subTrack[2] == "caption"}}
+                      {{captions = captions ? captions + ", " + subTrack[1] : subTrack[1]}}
+                  {{else}}
+                      {{subtitles = subtitles ? subtitles + ", " + subTrack[1] : subTrack[1]}}
+                  {{end}}
                 {{end}}
               {{end}}
             {{end}}
--- a/site/templates/film/item.jet
+++ b/site/templates/film/item.jet
@@ -92,14 +92,14 @@
             </div>
           {{end}}
           {{if len(film.GetSubtitles()) > 0 }}
-            {{subtitles:= ""}}
-            {{captions:= ""}}
-            {{range film.SubtitleTracks}}
-              {{if len(.) >= 3}}
-                {{if .[2] == "caption"}}
-                    {{captions = captions ? captions + ", " + .[1] : .[1]}}
+            {{subtitles := ""}}
+            {{captions := ""}}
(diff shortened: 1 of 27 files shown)


site/templates/film/item.jet
```diff
@@ -92,10 +92,29 @@
             </div>
           {{end}}
           {{if len(film.GetSubtitles()) > 0 }}
-            <div class="meta-detail-subtitle">
-              <h2>{{i18n("meta_detail_subtitles")}}</h2>
-              <p>{{ film.GetSubtitles() }}</p>
-            </div>
+            {{subtitles:= ""}}
+            {{captions:= ""}}
```
PeculiarGoat, line 96: Whitespace before colons please!

stajs:
I lol'd to see you broke convention on en_AU - I was expecting `Subtitles [CC]`! All good though :D

site/templates/film/item.jet
```diff
@@ -92,10 +92,29 @@
...
+            {{range film.SubtitleTracks}}
+              {{if len(.) >= 3}}
```
PeculiarGoat, line 98: I would prefer a named variable instead of `.`

PeculiarGoat (commented):
I wonder if it would be better to do this in Kibble: https://github.com/shift72/kibble/blob/99160804daeae0e84b7b97f9a2eb78899e909aaa/kibble/models/film.go#L94

Is there a reason for it being in core-template?

site/templates/film/item.jet
stajs: Agree, but it is idiomatic [[1](https://github.com/CloudyKit/jet/blob/master/docs/syntax.md#delimiters), [2](https://github.com/CloudyKit/jet/blob/master/docs/syntax.md#range)].

site/templates/film/item.jet
```diff
@@ -92,10 +92,29 @@
             </div>
           {{end}}
           {{if len(film.GetSubtitles()) > 0 }}
-            <div class="meta-detail-subtitle">
-              <h2>{{i18n("meta_detail_subtitles")}}</h2>
-              <p>{{ film.GetSubtitles() }}</p>
-            </div>
+            {{subtitles:= ""}}
+            {{captions:= ""}}
+            {{range film.SubtitleTracks}}
```
stajs, line 97: Based on my reading of [the changelog](https://github.com/shift72/kibble/blob/master/changelog.md#01014):

> Move field Subtitles to field SubtitleTracks. Add new field Subtitles which captures the likes of hard-coded subtitles. Add new function GetSubtitles() for tv episodes, films and bonus content. Populate this with a unique list of SubtitleTracks.Name and Subtitles values.

... you still need to range over `.GetSubtitles()` to ensure you cover both burnt and uploaded subtitles.

Potentially we might want to indicate burnt subs differently...?

stajs: And yeah, you should check the code is in line with that changelog too.

stajs (changes requested):
Just a question on covering burnt subs.

stajs:
> I wonder if it would be better to do this in Kibble: https://github.com/shift72/kibble/blob/99160804daeae0e84b7b97f9a2eb78899e909aaa/kibble/models/film.go#L94
> 
> Is there a reason for it being in core-template?

Well we needed the translations so a bit of a template change isn't a major extra. But I think the best fix is to stop lumping the captions together with subs so we don't have to do these maneuvers in kibble or core-template.

site/templates/film/item.jet
PeculiarGoat: Yes it is idiotic.

PeculiarGoat:
Should we investigate separating these at API level?

stajs:
Put the We in Web team.

site/templates/film/item.jet
brynyShift72: I don't think that hardcoded sub/CCs are returned by the api ? Unless I'm looking in the wrong place 😱 
I've added Czech as HC [to this film]( https://staging-test.shift72.com/admin/#/film/1619/subtitles)
but the film info has no mention of it when using [the api ](https://staging-test.shift72.com/services/meta/v2/film/1619/show_multiple)

stajs (approved):

PR #255, opened 2022-01-20
converted mvp sass variables to css variables

PeculiarGoat:
[Google Doc](https://docs.google.com/document/d/1H3zOyezfNvMCf_j5cJd-kq9Eu2Gxis5AIZw80KArBpE/edit#heading=h.2vddzvn6l12).

- If the site model has an external CSS link, add it in the head.
- Change SASS variables to CSS variables and replace usage in scss files.
- Develop workarounds for SASS and CSS functions that don’t understand CSS variables as arguments.

UPDATE 04/02/2022
- Added script which gets the latest custom css from live site (if there is any) and adds it to `site/styles/local.css` for use with local dev/testing.

UPDATE 09/02/2022
- Added script which lets you upload you local custom css file.


package.json
```diff
@@ -34,7 +34,7 @@
   "dependencies": {
     "@rollup/plugin-buble": "^0.21.3",
     "autoprefixer": "^7.1.2",
-    "bootstrap": "^4.0.0",
+    "bootstrap": "^4.6.1",
```
stajs, line 37: Assume this doesn't have Relish effects?

site/styles/_slider.scss
```diff
@@ -35,7 +35,10 @@ s72-userwishlist .s72-slider-control {
   &:hover,
   &:active,
   &:focus {
-    background: darken($primary, 7.5%);
+    // NOTE(jackH):
+    // no nice solution for darkening colour while using css variables
+    // commenting this out though because it doesn't seem to be in use anywhere
```
stajs, line 40: If not used, just del it I reckon.

site/styles/_variables.scss
```diff
@@ -36,76 +89,83 @@ $spacer: math.div(5, 8) * 1rem; // 10px
...
+  '4':
+    (
+      $spacer * 4,
+    ),
+  // 40px
+  '5':
+    (
+      $spacer * 5,
+    ),
+  // 50px
```
stajs, line 127: This looks like auto-formatting making it harder to read...

site/styles/_bootstrap4.scss
```diff
@@ -34,3 +34,93 @@
 @import 'bootstrap/scss/tooltip';
 @import 'bootstrap/scss/popover';
 @import 'bootstrap/scss/utilities';
+
+// The following override classes are to get around Bootstrap 4's use of Sass variables.
+// CSS themes will not work without these hacks.
+
+.btn-primary-override {
+  @extend .btn-primary;
+  border-color: $button-background;
```
stajs, line 43: How does this work? I can see the Sass var picking up from the CSS vars in `_variables.scss`:

```scss
--primary-rgb: 53, 160, 168;
--primary: rgb(var(--primary-rgb));
$button-background: var(--primary) !default;
border-color: $button-background;
```

So when the Sass is first compiled, it will have CSS for `border-color: var(--primary)`? Allowing override for either `--primary` or the underlying `--primary-rgb`?

stajs:
Looks good so far. Quite tricky but mostly a lot of search and replace?

package.json
PeculiarGoat: We were actually running `4.6.0` thanks to the `^`... I could probably change this back to `^4.0.0` ?

site/styles/_variables.scss
PeculiarGoat: Yeah I'll try to figure out how to fix that.

site/styles/_bootstrap4.scss
PeculiarGoat: - `--primary-rgb` is required for all the places where we use the `rgba` function.
- `--primary` is required for all the places where we (and Bootstrap) use colours directly without passing them through `rgba`.
- Those Bootstrap override classes are required for places where Bootstrap 'fails' silently while passing colours through functions.  In the case of the primary button, some colours would work and some would default back to Bootstrap colours.  I've extracted _all_ of the colours for the primary button however so it's easier to see the big picture in one place even though half of them don't actually need an override.

Does that make sense?

PeculiarGoat:
Lots of search and replace and super lots of picking through the Bootstrap repo to figure out what doesn't support CSS variables.

site/styles/_bootstrap4.scss
stajs: Yeah that makes sense, but wasn't what I was asking. I was asking for the compiled output in `main.css`, if the CSS in there was like this:

```css
/* Variables decalared early in the file */
--primary-rgb: 53, 160, 168;
--primary: rgb(var(--primary-rgb));

/* ... and much later when they are used (compiled from Sass - border-color: $button-background;) */
border-color: var(--primary);
```

And if so, then if the model was to have `local.css` be able to override the vars `--primary-rgb` or `--primary`?

PeculiarGoat: Yeah `local.css` can override any of the `--` vars.

stajs:
What's needed for this to not be draft?

PeculiarGoat:
@stajs potentially ready for review now.  I've just finished a CSS upload script.  But now wondering if it would be a good idea to change it around a bit so we can write Sass locally and have it compiled to CSS before uploading.  Thoughts?

stajs (approved):
Looks good! I think get this in as just CSS for now and see how it feels, if we want to do Sass later we can.

PR #266, opened 2022-02-14
added core-template version to head

PeculiarGoat:
- Added core-template version to `<head>`.  It's obtained via the site config model (kibble.json).
- Added node script which finds the core-template version in package.json and adds it to kibble.json.
- Added pre scripts to package.json to ensure that core-template version is added to kibble.json every time the site runs, the site is deployed, and core-template is published.

The kibble part of this task needs to be released before this can be merged.


l0ud0gg (approved):
LGTM

smitsanghvi21 (commented):
Seems to be working fine. But do we want to check if the specified `coreTemplateVersion` has been released? For ex, when I tried to manually change the version to `version": "0.4.9` in package.json and run `npm i`, it did try to install that version of core template.

Even though this doesn't exist. Though it doesn't break anything, but might cause confusion if another dev picks up this site and might wonder why the features of previous versions from 0.4.3 to 0.4.9 aren't working even if they are on 0.4.9. Thoughts @stajs @l0ud0gg ?

Screen Shot 2022-02-17 at 10 41 04 AM

package.json
```diff
@@ -21,7 +21,11 @@
     "kibble:build": "kibble publish",
     "build": "npm-run-all css js:build css:build kibble:build",
     "language:csv": "node scripts/language-file-to-csv/language-file-to-csv.js site/en_AU.all.json",
-    "language:validate": "node scripts/language-file-validator.js"
+    "language:validate": "node scripts/language-file-validator.js",
+    "version": "node scripts/core-template-version.js",
+    "prestart": "npm run version",
+    "prebuild": "npm run version",
```
stajs, line 27: I can't find this in [the lifecycle scripts](https://docs.npmjs.com/cli/v7/using-npm/scripts) - is `prebuild` legit?

package.json
```diff
@@ -21,7 +21,11 @@
...
+    "prepare": "npm run version"
```
stajs, line 28: Just noting this from the docs:

> As of npm@7 these scripts run in the background. To see the output, run with: --foreground-scripts.

Do you see output?

kibble.json
```diff
@@ -106,5 +184,6 @@
       "datasource": "Collection",
       "pageSize": 0
     }
-  ]
+  ],
+  "coreTemplateVersion": "0.4.2"
```
stajs, line 188: Is this PR the only reason why we add this to `kibble.json`? I'd prefer we right something to the [site JS dir](https://github.com/shift72/core-template/tree/AB%236706-add-core-template-version-to-head/site/static/js) to `console.log()` info than show in `<head>` so we don't have to add more things to `kibble.json`.

stajs (changes requested):
I'd prefer us to not add to `kibble.json` if we can avoid. Open to discussing. Is there another PR for shift72-template? I think these scripts should be there too?

package.json
PeculiarGoat: It's a [user-defined life cycle script](https://docs.npmjs.com/cli/v7/using-npm/scripts#npm-run-user-defined).

kibble.json
PeculiarGoat: There's a few places that need `coreTemplateVersion` from `kibble.json`:

- Local kibble uses it in local site builds to put in the <head>.
- Local kibble sends it as a param to the site templates end point during a deploy.  The site templates end point adds it to the site template database record here.
- S3 kibble uses it in the live site builds to put in the <head>.

Putting `coreTemplateVersion` in `kibble.json` kills all of those birds with one stone.  I had originally planned to add it as a command line arg for the publish cobra command in kibble but that is messier because the builder worker has to manually concatenate the command if the arg exists.

I think `kibble.json` is a fine place to put this because the template is the source of truth for this info (not the apis/db/whatever).

If you would prefer to `console.log` `coreTemplateVersion` instead of write it in the <head>, we can do that... but it still needs to go in `kibble.json` or through the publish cobra command to be passed to the site templates end point.

package.json
PeculiarGoat: I don't see anything out of the ordinary adding `--foreground-scripts` but I do see the logging from `core-template-version.js` so the `pre` scripts are working fine.

PeculiarGoat:
@stajs I haven't done this for shift72-template yet.  I suspect that it wont need anything changing because the kibble.json that it merges from core-template should already have the coreTemplateVersion property.

package.json
stajs: Can you verify you are on v7+?

kibble.json
stajs: What are the missing words at the end of these sentences?

> Local kibble uses it in local site builds to put in the .
> S3 kibble uses it in the live site builds to put in the .

This might win on easiness but it still feels that it pollutes kibble with a concept it shouldn't care about.

stajs:
> I haven't done this for shift72-template yet. I suspect that it wont need anything changing because the kibble.json that it merges from core-template should already have the coreTemplateVersion property.

Won't shift72-template (and the resulting sites) need to be running their own lifecycle scripts in order to update the core version?

package.json
PeculiarGoat: I'm on 8.1.1

kibble.json
PeculiarGoat: Damn what the?!?

PeculiarGoat: Huh... I guess Github was stripping out my unescaped `<head>` tags lol.  Edited the comment above!

PeculiarGoat:
shift72-template merges `local/kibble.json` and `./node_modules/@shift72/core-template/kibble.json` together on init.  As long as `local/kibble.json` has no coreTemplateVersion entry, the output should always have coreTemplateVersion from core-template.... which means that in theory, shift72-template should magically get this feature for free when it upgrades core-template to a release containing this feature.

kibble.json
```diff
@@ -5,32 +5,110 @@
...
-    "nl": { "code": "nl_BE", "name": "Nederlands" },
-    "no": { "code": "no_NO", "name": "Norsk" },
-    "pl": { "code": "pl_PL", "name": "Polski" },
-    "pt": { "code": "pt_PT", "name": "Português" },
-    "pt-br": { "code": "pt_BR", "name": "Português do Brasil" },
-    "ru": { "code": "ru_RU", "name": "Pусский" },
-    "se": { "code": "se_SE", "name": "Српски" },
-    "tr": { "code": "tr_TR", "name": "Türk" },
-    "uk": { "code": "uk_UA", "name": "Український" },
-    "zh-tw": { "code": "zh_TW", "name": "中文" }
```
stajs, line 33: Can we revert the newlines?

stajs (approved):
:shipit:

stajs:
@PeculiarGoat adding a blocked label for now - did the kibble release last week satisfy the requirement?

PeculiarGoat:
@stajs the last kibble release was 17 days ago so no.

kibble.json
PeculiarGoat: No... it's a side effect of the file being modified.  How often do you need to look at the stuff below the languages though?

stajs: Not a lot. Hopefully we get the languages/translations out of here soon enough.

PR #270, opened 2022-02-27
Update language files

brynyShift72:
Added missing translations to each language (missing keys are using the English translations ), removed duplicate translations


PeculiarGoat:
The validator still warns that they have english values in non-english files right?

PeculiarGoat (approved):

PR #274, opened 2022-03-03
primary-scss-to-css-variable

brynyShift72:
fixin' $primary

### DIFF
--- a/site/styles/_buttons.scss
+++ b/site/styles/_buttons.scss
@@ -115,8 +115,8 @@
     background: transparent;
 
     .s72-icon-play circle {
-      fill: $primary;
-      stroke: $primary;
+      fill: var(--primary);
+      stroke: var(--primary);
       stroke-width: 3px;
     }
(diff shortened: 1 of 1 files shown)


PeculiarGoat (approved):

site/styles/_buttons.scss
```diff
@@ -115,8 +115,8 @@
     background: transparent;
 
     .s72-icon-play circle {
-      fill: $primary;
-      stroke: $primary;
+      fill: var(--primary);
```
l0ud0gg, line 118: should these not be `var(--primary-rgb)`?

l0ud0gg: Ah --primary is set based off --primary-rgb `--primary: rgb(var(--primary-rgb));`

l0ud0gg (approved):
looks good. lets merge

smitsanghvi21 (approved):

PR #276, opened 2022-03-03
AB#6926 Move CTA button for plans

additiverich:
[ADO](https://dev.azure.com/S72/SHIFT72/_boards/board/t/Web%20team/Stories/?workitem=6926) | [Zeplin ](https://app.zeplin.io/project/607f961cf9f03790cc66e475/screen/61396196750e52b971758c75)

- move CTA button into pageHeader
- pageContent doesn't need a plan anymore
- modify page header CSS for curated plan show case pages only
- create 1rem spacing variable for vertical spacing purposes


site/templates/page/page-header.jet
```diff
@@ -5,9 +5,9 @@
         <s72-plan-label data-slug="{{plan.Slug}}" data-title="{{plan.Name}}"></s72-plan-label>
       {{end}}
       <h1>{{title}}</h1>
-      {{if plan != 0 && plan.HasExpiryDate() }}
-        {{ if plan.HasExpiryDate() }}
-          <p class="alert p-0 border-0 my-1">
```
matthewthomsonnz, line 10: My opinionated take is this is a step in the wrong direction

site/styles/_variables.scss
```diff
@@ -105,6 +105,7 @@ $spacers: (
   // 40px
   '5': $spacer * 5,
   // 50px
+  '1rem': 1rem
```
stajs, line 108: lel

site/templates/page/page-header.jet
stajs: How so?

stajs (approved):
Looks fine to me, _however_  just to throw an idea: is it worth having a specific page type for plans rather than overloading curated pages?

site/templates/page/page-header.jet
additiverich: Ping @matthewthomsonnz - what's the step in the wrong direction?

matthewthomsonnz:
```
.curated-plan-page {
  .page-header {
    .alert {
      @extend .p-0;
      @extend .border-0;
```
I know inline utility classes are controversial, but I think they are a good solution for one offs, and mitigate nested css hell

stajs: Ah, the removal of the utility classes are the step in the wrong direction. I'm not against utility classes per-se and `@extend` can lead to more bloat in the final CSS. In this case I agree with you, but it's pretty marginal so I don't mind either way.

stajs: PS, reason why it was hard to spot the problem, is the view on the "Files changed" tab points out the utility classes while the one in "Conversation" doesn't:

image

image

additiverich: @matthewthomsonnz Interested in your feedback now

site/templates/page/page-header.jet
```diff
@@ -1,13 +1,13 @@
...
   {{if len(title) > 0}}
-    <div class="page-header">
+    <div class="page-header pb-0">
       {{if plan != 0}}
         <s72-plan-label data-slug="{{plan.Slug}}" data-title="{{plan.Name}}"></s72-plan-label>
       {{end}}
       <h1>{{title}}</h1>
-      {{if plan != 0 && plan.HasExpiryDate() }}
-        {{ if plan.HasExpiryDate() }}
-          <p class="alert p-0 border-0 my-1">
```
matthewthomsonnz, line 10: Re-add back in the 'alert' inline class and good to go (without it, the icon color rule [here](https://github.com/shift72/core-template/blob/develop/site/styles/_icons.scss#L35-L37) is not primary as per [design](https://app.zeplin.io/project/607f961cf9f03790cc66e475/screen/61396196750e52b971758c75))

PR #280, opened 2022-03-04
Added skip link

curtisadams28:
Deployed to [staging-test.shift72.com](https://staging-test.shift72.com/)

The carousel keyboard control is broken so ignore that. You can see it working properly on a film page.


stajs:
1. Only problem is the cookie consent means the skip link isn't the first tab - can we move it above the cookie consent in tab order?
2. Just a marginal thing - on a dark theme or black image, it's not terribly visible
3. Should the page move down at _all_ when the link is activated? It seems the Microsoft moves a smidge, but ours doesn't (tested on a page with lots of content so there is scrollbars).

curtisadams28:
@stajs 
1. Done.

2. I can change it to body-bg-accent which is slightly lighter.
image
image
Alternatively, we could use body-color as the background.
image

3. It skips to #main which currently sits at the top of the page. Once the navigation is relatively positioned (another one), it should skip downwards.

stajs (commented):
:shipit:

stajs (approved):

PR #283, opened 2022-03-07
i18next

PeculiarGoat:
FOR DEMONSTRATION ONLY.  DO NOT MERGE!


site/static/js/i18next-test.js
```diff
@@ -0,0 +1,125 @@
...
+    ar	  test_pluralities    100	   I do other plurality!
+  */
+
+  /*
+    Notes:
+
+    Translations that use pluralities (like all of ours) need to be reformatted like so:
+      "some_key_one": "some value",
+      "some_key_other": "some values",
+      etc...
```
stajs, line 116: Hmm, it seems that we would need to re-format the JSON to the schema required by this library - that's a bit balls but could be done. I think it would only be necessary for ones using a count, so wouldn't be all of them.

PeculiarGoat: Yeah.  The processTranslations function on line 22 takes care of this okay... but it would be nicer if we didn't have to bother at all.

brynyShift72:
Is all this getting moved over to relish ?

stajs:
@brynyShift72 Yeah, I think @PeculiarGoat just put it here for easier demonstration purposes. Not that this is really secret stuff but we'll have to be careful about what gets shared like this PR in the public core-template repo versus the private relish repo.

PR #286, opened 2022-03-11
Idea for admin start

stajs:
Just an idea, instead of having these scripts being called like:

```
npm start
npm run start-admin
```

... we could use arguments like this:

```
npm start
npm start --admin
```

P.S. `run-p` is shorthand for `npm-run-all --parallel`.

I added a `kibble:start` that always `mkdir`'s (why not?) and tests for the `--admin` flag and chooses either `kibble:watch` or `kibble:watch-admin`. I only broke it out this way so I could do some fancy schmancy console coloring, but `kibble:watch-admin` is simply calling back to `kibble:watch` and passing in it's own `--admin` flag. So `kibble:start` could be just:

```
... npm run kibble:watch -- --admin || npm run kibble:watch
```

---

image
image


stajs:
Speaking of colors, we should colorize that `api key has expired` - I'm always getting caught out sat waiting for things to run when that scrolls past and don't see it.

brynyShift72:
I like this way better just wondering:
Can you help me understand the `&& test -n \"$npm_config_admin\"` I assume that this is checking for the --admin argument some how?

brynyShift72:
I like this way better just wondering:
Can you help me understand the `&& test -n \"$npm_config_admin\"` I assume that this is checking for the --admin argument some how?

stajs:
- The flag in `npm start --admin`: [a flag without a value gets set to true](https://docs.npmjs.com/cli/v7/using-npm/config#command-line-flags) and then is [available as an environment variable](https://docs.npmjs.com/cli/v7/using-npm/config#environment-variables) which is snake_cased and prefixed with `npm_config_` so `--admin` becomes `$npm_config_admin`. If the flag is not specified then this environment variable is empty.
- Wrapping in double-quotes to turn it to a string `\"$npm_config_admin\"`.
- `test` is a shorter syntax than an if/else block and `-n` is true if the string is non-zero length.

brynyShift72:
Sweet, good to know : )

PeculiarGoat (approved):
LGTM

brynyShift72 (approved):

l0ud0gg (approved):

smitsanghvi21 (approved):

stajs:
Converted usages (where possible) to shorthand versions of `npm-run-all` to allow logging output of task name.

PR #288, opened 2022-03-17
Implement the designs 6846

matthewthomsonnz:


stajs:
@matthewthomsonnz I know this is still draft, but this looks pretty meaty and I'm a bit worried about what's in here - can we get another joint session to discuss?

@curtisadams28 fyi some more carousel changes.

matthewthomsonnz:
> @matthewthomsonnz I know this is still draft, but this looks pretty meaty and I'm a bit worried about what's in here - can we get another joint session to discuss?

@stajs it sure is, and definately understandable. Can do, do you want to set one up? I'm free any time, today even

PR #290, opened 2022-03-21
Ab#6003 carousel keyboard control

PeculiarGoat:
- Removed existing broken focus ring from `.carousel-item-link`.
- Added focus ring to `.carousel-item-caption` when `.carousel-item-link` is focused.
- Changed pagination dots to button elements so they can be focused.
- Added aria-label to pagination dots with translations for screen readers.


PeculiarGoat:
image
image

PR #291, opened 2022-03-22
Ab#5032 use image map for films

additiverich:
- upgrade to Kibble 0.16.6, which introduces ImageMap for film objects (in parallel with Images but does not care what they are called)
- references to film.Images changed to film.ImageMap
- update changelog


stajs (approved):
LGTM, but surprised it only touches the film model - no images for TV/pages?

additiverich:
> LGTM, but surprised it only touches the film model - no images for TV/pages?

Not yet - we're limiting these changes to just films (with a view to doing the rest later)

PR #292, opened 2022-03-23
Implement the designs - availability label

matthewthomsonnz:
- This PR moves and styles the availability label in carousel, film detail, and tv detail pages as per latest designs
- It also moves and styles it in the bottom of each slider poster

[Designs link](https://app.zeplin.io/project/5fdaa165fbacf584a2ce4f8b/dashboard)

As discussed this PR is a small chunk of a larger PR in order to make things easier to review

[AB#6846](https://dev.azure.com/S72/fefacba9-96b6-4af2-a53d-050ed453e13a/_workitems/edit/6846)


stajs:
FYI, I've replaced @smitsanghvi21 with @curtisadams28 for review.

curtisadams28:
@matthewthomsonnz The film detail/carouel looks great but the availability label on sliders isn't very readable.
1. The text extends all the way to the edges.
2. The background opacity is low which makes it low contrast for white posters.
3. The capitalisation means there is less contrast in the size and shape of letters. I think the availablity label is too close to a sentence to get away with uppercase. Especially when you factor in translations.

Might be worth showing design the screenshots below:
Screenshot 2022-03-24 145824 Screenshot 2022-03-24 145803

stajs:
Just on the uppercasing, I think the reorganizing has set it free - it was uppercase but reset later, e.g. TVOD demo site:

image

I'm opposed to us doing any uppercasing on translations. My first thought is [the Turkish problem](https://stackoverflow.com/questions/3777443/css-text-transform-not-working-properly-for-turkish-characters), but moreso we should be respecting our clients translations verbatim - if they wanted uppercase they can choose to do so.

---

Leading on, I'm also opposed to uppercasing in general. We have a number of client requests to change it on the tagline which currently means a customization. For example, this is a directors name they don't like uppercased (`Susan Sontag`) and they don't like the hours/minutes either (`1h 37m`).

image

matthewthomsonnz:
@curtisadams28 Wow that screenshot made me realise text is much lighter weight on Ubuntu :S

- Have added 5px padding on left/right
- Have upped the bg opacity from 0.5 to 0.8
- Have dropped the weight to 400 - @curtisadams28 do you reckon this helps contrast in the size and shape of letters a bit?
- Cool I'll show design some before after (windows) screenshots

@stajs  I'll ask Design/Product how they want to proceed with the capitalisation. Is it only certain countries/languages with this bugbear? Could you see a compromise as something like the following?
```
 .selector:lang(en,es ..etc) { text-transform: uppercase; }
or
 .selector:not(lang(tr ..etc)) { text-transform: uppercase; }
```

stajs:
It's not certain countries/languages - I'm pretty adamant that we don't transform a clients translations.

FYI, @curtisadams28's screenshot will be Windows.

curtisadams28:
@matthewthomsonnz The contrast is much better now 👍

site/styles/_availability-tags.scss
```diff
@@ -1,43 +1,63 @@
...
-    margin-left: auto;
-    margin-right: auto;
-    padding: 10px 10px 5px;
-    right: 0;
+    position: absolute;
+    z-index: 99;
+  }
+
+  s72-availability-label {
+    background-color: rgba(var(--body-bg-rgb),0.7);
```
PeculiarGoat, line 29: Shouldn't this be on the `.availability-bg` element?

site/styles/_availability-tags.scss
```diff
@@ -1,43 +1,63 @@
...
-      text-align: center;
+      line-height: 18px;
+      @extend .mb-0;
+      &:before {
+        content: none;
+      }
     }
   }
+
+  s72-availability-status .availability-bg {
```
PeculiarGoat, line 44: image

This is different to the designs:
image

It's also broken on the detail page:
image

site/styles/_availability-tags.scss
PeculiarGoat: Also... your comment said you put it back to 0.8 opacity... was that a typo?

site/styles/_meta-item.scss
```diff
@@ -98,3 +98,11 @@
 .meta-item-buttons {
   margin-top: 5px;
 }
+
+// both meta detail and carousel
+.meta-item-tagline {
```
PeculiarGoat, line 103: This is overridden by `.carousel-item-caption .meta-item-content .meta-item-tagline` in the carousel and by `.meta-detail-content .meta-item-tagline` on the detail page.

PeculiarGoat (changes requested):
Some things to looks at.  Could you please run the linter too `npm run css:lint`.

site/styles/_availability-tags.scss
matthewthomsonnz: - Ah yes .availability-bg makes more sense. Changed.
- Yeah 0.8 opacity was a typo. I think i set it to 0.8, then realised 0.7 was a good medium when the design wanted 0.5

site/styles/_availability-tags.scss
matthewthomsonnz: I'm trying to feed the overall changes in incrementally, but I think you're right, it should look like the design as I've touched it in this PR, therefore:

- Have matched it to the design now
- This PR now also updates the sitewide button bg/text color (and this label now inherits those colors)
-  That broken detail page appears in develop branch, do you reckon we should fix it though? As the label in screenshot is artificially long. If so is it cool if I fix it in an upcoming future branch (It updates the detail page poster widths to the designs)?

(Context: This is [the branch](https://github.com/shift72/core-template/tree/Implement-the-rest-of-the-designs-6846) discussed in our interteam collab with all changes (Feeding in as smaller PRs))

site/styles/_meta-item.scss
matthewthomsonnz: Removed

matthewthomsonnz:
> Some things to looks at. Could you please run the linter too `npm run css:lint`.
- css linted

matthewthomsonnz:
> It's not certain countries/languages - I'm pretty adamant that we don't transform a clients translations.
> 
> FYI, @curtisadams28's screenshot will be Windows.

@stajs discussed with peeps and we will hold off capitalisation transform for now. Will add to agenda next inter-team meeting.

site/styles/_availability-tags.scss
PeculiarGoat: IMO it's worth fixing now... clients can put whatever they want in those translations so I think we should be handling extra long labels.

matthewthomsonnz: Oh yes good point. Fix now applied.

PeculiarGoat: Much better!
image

site/styles/_availability-tags.scss
```diff
@@ -1,43 +1,67 @@
...
+  .availability-expired {
+    @extend .d-none;
+  }
+
+  .availability-state {
+    @extend .justify-content-center;
+    color: var(--body-color-rgb);
+    font-size: 12px;
+    text-align: center;
+    text-transform: uppercase;
```
PeculiarGoat, line 65: So we are decided on not doing this transform right?

Something worth noting is that even if the client makes these labels all-caps via the translations, dates coming from Relish will still have lowercase characters:
image

So maybe we need to give this some extra thought.

PeculiarGoat:
@matthewthomsonnz are you still adding to this?

matthewthomsonnz:
> @matthewthomsonnz are you still adding to this?

Nope! All finished now.

I realised the label style was also used on plans.html - have now applied styling to them also.

PeculiarGoat (approved):
I'm happy if you are @matthewthomsonnz !  :+1:

stajs:
@l0ud0gg @curtisadams28 Can you complete reviews on this too and then I think we'll take it over in @matthewthomsonnz's absence.

additiverich:
@stajs I've moved the task which was tracking this into the delivery team board

stajs:
@additiverich no biggie but can you land it in our parking lot for future tho, cheers. It'll end up in our sprint anyway but just gives us a chance to sight incoming work.

PR #294, opened 2022-03-23
automated translate

matthewthomsonnz:
Hey guys. I just spent an hour hacking this together this morning. try it out!
e.g.
edit existing key: `npm run translate play_button_watch bananas`
new key: `npm run translate some_new_key apples`

as a bonus formats everything consistently to two indents


stajs (approved):
I haven't tested but LGTM.

- Since it's kinda incidental it could be merged and tested as we go.
- I notice the choice of Michael Jackson Script over normal `.js` - there is probably a discussion to be had on relative merits and consistency.
- I've raised the Estonia mislabelling as a bug to fix.
- This might be a limited life thing with the move to translations in the database, although I'd love the concept to persist, just a matter of working out the process.

@l0ud0gg

translate.mjs
```diff
@@ -0,0 +1,37 @@
+import fetch from 'node-fetch';
+import glob from 'glob';
+import * as fs from 'fs';
+const key = process.argv[2];
+const value = process.argv[3];
+
+if (!key || !value) {
+    console.log("Should be 'npm run translate key value'")
+    process.exit(1)
```
PeculiarGoat, line 9: Missing semi-colons.  `npm run js:lint` will catch these.

translate.mjs
```diff
@@ -0,0 +1,37 @@
+import fetch from 'node-fetch';
```
PeculiarGoat, line 1: @stajs IMO the Michael Jackson Script imports are nicer and we should swap out those common JS imports in the other scripts.

translate.mjs
```diff
@@ -0,0 +1,37 @@
...
+if (!key || !value) {
+    console.log("Should be 'npm run translate key value'")
+    process.exit(1)
+}
+glob('./site/*.json', {}, (err, files) => {
+  files.forEach(pathName => {
+    const file_content = fs.readFileSync(pathName);
+    const content = JSON.parse(file_content);
+    let languageCode = pathName.slice(7, 9);
+    if (languageCode == 'ee') languageCode = 'et';
```
PeculiarGoat, line 16: Better to fix the dodgy language code than add hacks to sidestep it!

translate.mjs
```diff
@@ -0,0 +1,37 @@
...
+        const contentType = res.headers.get('content-type');
+        if (contentType && contentType.indexOf('application/json') !== -1) {
+          return res.json();
+        }
+        console.log(`translation failed for ${languageCode}`);
+      })
+      .then(translated => {
+        if (translated) {
+          const translatedValue = translated[0][0][0];
+          content[key] = { other: translatedValue };
```
PeculiarGoat, line 32: What about the other pluralities?  Maybe the script can take an extra arg to tell it which plurality to add.

package-lock.json
```diff
@@ -1,8098 +1,8 @@
 {
   "name": "@shift72/core-template",
   "version": "0.4.4",
-  "lockfileVersion": 2,
+  "lockfileVersion": 1,
```
PeculiarGoat, line 4: Time to update npm!

PeculiarGoat (changes requested):
This is cool!  Few things to look at though.

translate.mjs
PeculiarGoat: We can sneak it into this PR though right?

PeculiarGoat:
@matthewthomsonnz I forgot to add that this script can go in the `scripts` directory!

translate.mjs
stajs: If you want to get in on the PR sure, but I don't expect him to clean up our mess.

package.json
```diff
@@ -105,6 +113,7 @@
   },
   "eslintIgnore": [
     "modernizr-custom.js",
+    "language-file-to-csv.js",
```
PeculiarGoat, line 116: Why ignore this?

matthewthomsonnz: Since adding that package-lock I keep getting [this error](https://github.com/shift72/core-template/runs/5730066747?check_suite_focus=true). Any ideas? I tried for a long time but no success adding the prettier ignore option.

PeculiarGoat: I've just pushed a fix into develop so you can merge that in and see if it it's okay... strange that running the linter locally doesn't pick that up but running in the Github Action does.  Also strange that we haven't noticed it before!

matthewthomsonnz: Oh sweet. Nice one. Merged develop into this branch and reverted ignore

PeculiarGoat: Do you need to change this:
`"js:lint": "prettier '{scripts,site/static/js}/**/!(modernizr-custom).js' --write && eslint ./site/static/js/*.js ./scripts --fix"`

to this:
`"js:lint": "prettier '{scripts,site/static/js}/**/!(modernizr-custom).*js' --write && eslint ./site/static/js/*.js ./scripts --fix"`

so the linter will pick up that mjs file

matthewthomsonnz: Ah yep that was a good idea. Cheers added.

PeculiarGoat (approved):
Nice work!

curtisadams28 (approved):

PR #295, opened 2022-03-23
Update 6 different translations programmatically

matthewthomsonnz:
Hey guys,

I was being asked to update 6 different translations in some card feedback ([#AB6477](https://dev.azure.com/S72/SHIFT72/_boards/board/t/Web%20team/Stories/?workitem=6477)), I realised it was quicker and better for morale to figure out how to just automate a translation, than to actually do it 6 times over manually.

So I did that. [Here is the PR with the script ](https://github.com/shift72/core-template/pull/294)

The purpose of this PR ensure you guys are okay with the result of this, including indents changing and losing any newlines

- (Note: This is a pull request into the share modal branch)

### DIFF
--- a/site/ar_LB.all.json
+++ b/site/ar_LB.all.json
@@ -1,595 +1,1260 @@
 {
-    "site_owner": { "other": "ABC Cinemas" },
-    "settings_title": { "other": "إعدادات" },
-    "powered_by": { "other": "مشغل بواسطة" },
-    "powered_by_name": { "other": "Shift72" },
-    "powered_by_url": { "other": "https://www.shift72.com" },
-    "in_partnership_with": { "other": "بالشراكة مع/ بالتعاون مع" },
-    "festival_scope": { "other": "Festival Scope" },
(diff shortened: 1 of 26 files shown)


stajs:
Unfortunately the GitHub diffing of large changes here makes it near impossible to comprehend. In general I'm okay with autoformating from:

```json
"key": { "one": "tahi", "other": "atu"}
```

... to:

```json
"key": {
  "one": "tahi",
  "other": "atu"
}
```

So in the interests of sanity can I request a PR to _just_ do the auto-formatting that we can blindly accept on faith _and then_ a PR just for the real changes?

stajs:
Oh yes, and that auto-formatting PR, if we can get that merged to all active branches at round the same time would be ace to avoid the diffing issue across them all.

matthewthomsonnz:
> So in the interests of sanity can I request a PR to _just_ do the auto-formatting that we can blindly accept on faith _and then_ a PR just for the real changes?

[👍](https://emojipedia.org/thumbs-up/) I'll close this PR.

Once translate script is in develop I'll make the auto-formatting-only PR. Then when thats in develop will get it merged into open branches.

PR #297, opened 2022-03-28
Allow overriding of the cookie consent variables

stajs:
1. The `$cookie-consent-*` vars were missing `!default` so couldn't override them.
2. Reformatted a bit of the changelog.


PeculiarGoat (approved):
:+1:

PR #298, opened 2022-03-28
Ab#6674 sub self service

ashesnz:
[ADO Card](https://dev.azure.com/S72/SHIFT72/_boards/board/t/Web%20team/Stories/?workitem=6674)

This is PR is dependant on the following [Relish PR](https://github.com/indiereign/shift72-web/pull/261)


site/en_AU.all.json
```diff
@@ -433,7 +461,7 @@
   "shopping_card_new_subscription": { "other": "This card will be saved and charged regularly" },
   "shopping_card_change": { "other": "Credit card not correct? <a href=\"{{.SubscriptionsURL}}\">Click here to update your card.</a>" },
 
-  "shopping_info_ownership_plan": { "other": "Subscription"},
+  "shopping_info_ownership_plan": { "other": "shopping_info_ownership_plan" },
```
additiverich, line 464: What's going on here?

ashesnz: Hmmm, nice observation. Wonder if I did a merge of my local develop branch rather than remote

additiverich: I noticed a few translations like that

stajs:
We'll need the changelog and other language files updated too - @matthewthomsonnz added a handy-dandy auto-translate script, see #294 but we need #300 merged first to avoid major conflicts on formatting.

ashesnz:
> > > We'll need the changelog and other language files updated too - @matthewthomsonnz added a handy-dandy auto-translate script, see #294 but we need #300 merged first to avoid major conflicts on formatting.
> > 
> > 
> > Thanks for that @stajs. Just a quick question, If you have the following translation string
> > ```
> > "usersubscriptions_subscriptions_section_content": {
> >   "other": ""
> > },
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > is that the same as not putting the translation string in the file at all. Even if it is the same, do you guys still prefer showing that in all translation files, or not really fussed?
> 
> We're planning on moving all of these translations to the database, so its best to have all keys in each language file (even if they are empty), however, it would probably be best to have at least some translation in there, as in the future a blank translation will stop the value from being rendered, which might be a bit finicky to track down.

Thanks @brynyShift72. Ive just done a [quick hack](https://github.com/shift72/core-template/pull/304) so we are able to handle the case(s) you have described. As I noticed I couldn't convert a few translations

ashesnz:
> We'll need the changelog and other language files updated too - @matthewthomsonnz added a handy-dandy auto-translate script, see #294 but we need #300 merged first to avoid major conflicts on formatting.

Added further (international) translations and dried up that auto translate file at the same time.

> We'll need the changelog and other language files updated too - @matthewthomsonnz added a handy-dandy auto-translate script, see #294 but we need #300 merged first to avoid major conflicts on formatting.

Updated changelog, added further user subscriptions translations and also dried up that auto translate file.

site/styles/_buttons.scss
```diff
@@ -200,6 +200,26 @@
...
+
+.s72-btn.s72-btn-explore-subs {
+  background: #1aa3ff;
+  border: 3px solid #1aa3ff;
+  min-width: 140px;
+  @include hover-focus {
+    background: #55bbff;
+    border-color: #55bbff;
+  }
+}
```
additiverich, line 222: This makes me go "hmm" - are these colours only actually used on the subs page? Or do we need to use colours defined elsewhere? This might be a question to go back to Design on.

site/styles/_pages.scss
```diff
@@ -33,6 +33,18 @@
...
+    font-weight: $font-weight-bold;
+    margin-bottom: 0;
+    @include media-breakpoint-up(md) {
+      font-size: 36px;
+    }
+    @include media-breakpoint-up(xl) {
+      font-size: 38px;
+    }
+  }
+
```
additiverich, line 47: If I'm reading this right, h2 defaults to larger bigger than h1 unless it's a desktop view, in which case they're the same? And our h3 is the same as h1 (20px) on the mobile view?

additiverich (commented):
A couple of questions from someone who is trying to approve CSS by looking at the declarations rather than in a browser.

site/styles/_pages.scss
ashesnz: Yeah I was a bit confused by these as well as I just followed the styleguide settings provided in Zeplin, I'll raise these issues with Bex, didn't realize they were inconsistent throughtout.

image

ashesnz: Reverted CSS changes. Will look at creating another PR for just the design aspect.

ashesnz:
@stajs Are we able to have a look at this? (or someone in delivery) to get approval (or not). Just wanting to get this ready for testing. Much appreciated.

additiverich (approved):

stajs:
@l0ud0gg @curtisadams28 @brynyShift72 y'all need to keep your review commitments plz

brynyShift72 (approved):
LGTM! 
As French and English are both in the database you may want to upload those as well, there is a wiki about that process [here](https://dev.azure.com/S72/SHIFT72/_wiki/wikis/SHIFT72.wiki/214/Self-Service-Translation-update-process) , else we can update at a later time

stajs:
Let's keep the database stuff out for now, until we've had a good run through.

site/templates/application/nav/user_logged_in.jet
```diff
@@ -26,6 +26,11 @@
         <li>
           <a class="s72-dropdown-item" href="{{ routeToPath("/devices.html") }}">{{i18n("nav_devices")}}</a>
         </li>
+        {{if config("account_show_plans") == "true" }}
```
l0ud0gg, line 29: Looks like this config is enabled for around 65 clients at the moment so i guess we may want to consider if it should show the subscription page for all use cases. (i think most of those clients are VIP sites where they are invited to plans rather than the user signing up / subscribing so could be wierd in terms of context for the user).

site/styles/_buttons.scss
l0ud0gg: +1

l0ud0gg (changes requested):
prolly worth just checking if we want to enable the subscription page for existing clients that have the account_show_plans config enabled in the VIP / invite to plan context. I just added a +1 to DR's comment on hardcoded colours. if its suitable then it would be good to use variables to keep in line with theming, otherwise all goods.

site/styles/_buttons.scss
ashesnz: Sorry should of marked this convo as resolved. This was updated recently, hence the outdated file.

site/templates/application/nav/user_logged_in.jet
ashesnz: Thanks for the heads up. I will check this up with either Jono or Nev

ashesnz: @l0ud0gg Thanks for that feedback. We've decided to go for a completely new config `account_show_subscriptions` instead of `account_show_plans`

ashesnz:
> prolly worth just checking if we want to enable the subscription page for existing clients that have the account_show_plans config enabled in the VIP / invite to plan context. I just added a +1 to DR's comment on hardcoded colours. if its suitable then it would be good to use variables to keep in line with theming, otherwise all goods.

Sorry may need to revert those changes, Jono just some further clarification

l0ud0gg (approved):
As discussed, if you revert changes back to account_show_plans and the audit is done to remove it from existing sites its all good :) I've updated documentation around the toggle for future reference.

PR #299, opened 2022-03-29
AB#6719 support ga ua legacy and ga4

l0ud0gg:
Goal of this is to support Google Analytics 4 if thats the type of tracking code the client has provided. In admin the admin user can add a GA tracking code - at the moment our core template code can only ulitise a legacy UA prefixed tracking code. These changes check if its the newer G prefixed GA4 tracking code and if so loads the necessary script and code to log tracking to GA (before these changes it fails silently does nothing). It also still honors the cookie acceptance. It will only create a cookie / talk to GA if the cookie policy has been accepted.

Note: the ecommerce support that is in relish still relies on UA legacy GA if tag manager isnt setup. Im going to create a card and treat as a separate issue since its not working at the moment anyway. This change is purely for traffic tracking.

### DIFF
--- a/site/templates/application/google.jet
+++ b/site/templates/application/google.jet
@@ -1,5 +1,9 @@
 {{block googleScripts(tagManagerId=config("google_tag_manager_id"), analyticsId=config("google_analytics_id"))}}
   <!-- Google integration scripts -->
+  <!-- GA4 script if G tracking code is provided -->
+  {{ if hasPrefix(analyticsId,"G")}}
+    <script async src="https://www.googletagmanager.com/gtag/js?id={{analyticsId}}"></script>
+  {{ end }}
   <script>
     function loadGoogleTagManager() {
       (function(w,d,s,l,i){w[l]=w[l]||[];w[l].push({'gtm.start':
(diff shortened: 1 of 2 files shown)


site/templates/application/google.jet
```diff
@@ -1,3 +1,9 @@
+{{block googleGa4Script(analyticsId=config("google_analytics_id"))}}
+  {{if hasPrefix(analyticsId,"G")}}
+  <!-- GA4 script if G tracking code is provided -->
+    <script async src="https://www.googletagmanager.com/gtag/js?id={{analyticsId}}"></script>
+  {{end}}
+{{end}}
```
PeculiarGoat, line 6:
```suggestion
{{end}}

```

PeculiarGoat (approved):
LGTM :+1:

stajs (approved):
Code LGTM, but I have a question regarding consent. `UA` would dynamically create the `<script>` based on any consent conditions, but now `GA4` will always be statically including its `<script>` in the `<head>`. I'm not sure if the new method has a dynamic option but it's probably worth checking to see if we can keep that existing behaviour as it feels more "privacy-first" approach we should be taking if possible.

l0ud0gg:
> 
yeah i did have a look into it. I couldnt find any snippets to dynamically write the script. I did test that loading the script doesnt set any cookies or send any data until consent is given. that being said i agree but i dont know how to achieve it.

PR #300, opened 2022-03-29
format translations for auto translate script

matthewthomsonnz:
TLDR: this PR just re-indents all the language json files

- We have now [merged a script into core-template](https://github.com/shift72/core-template/pull/294) that automatically translates a key for each translation file
- A side effect of this was all translations had to be auto formatted as it [rewrites the JSON with two indents](https://github.com/shift72/core-template/blob/develop/scripts/translate.mjs#L39)
- So @stajs [requested a PR](https://github.com/shift72/core-template/pull/295#issuecomment-1076942648) that does the formatting in one fell swoop
- Should theoretically be no different apart from intents/newlines, but cast your eyes over it just in case

### DIFF
--- a/site/ar_LB.all.json
+++ b/site/ar_LB.all.json
@@ -1,592 +1,1257 @@
 {
-    "site_owner": { "other": "ABC Cinemas" },
-    "settings_title": { "other": "إعدادات" },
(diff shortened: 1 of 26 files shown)


stajs (commented):
Happy to merge on faith once the conflict is resolved.

PR #306, opened 2022-04-04
Pull out yield / block for inserting G4 script into application.jet a…

l0ud0gg:
…nd add anonymous function in the loadGoogleAnalytics function to create the async script tag for us after cookie consent is confirmed.


stajs:
Ready for review?

l0ud0gg:
> 

was gonna do a bit more testing first.

PR #312, opened 2022-04-06
Ab#6719   load ga4 script dynamically

l0ud0gg:
Ok, so ive removed the block / yield to load the GA4 async script and wrote it as an anonymous function instead, so its only loaded once the cookie consent has been checked. Ive tested it locally thoroughly, switching between Legacy GA and GA4 to make sure data is collected and only if cookie consent has been given. I've also tested that the GA4 code works alongside google tag manager.

### DIFF
--- a/site/templates/application/google.jet
+++ b/site/templates/application/google.jet
@@ -1,24 +1,32 @@
-{{block googleGa4Script(analyticsId=config("google_analytics_id"))}}
-  {{if hasPrefix(analyticsId,"G")}}
-  <!-- GA4 script if G tracking code is provided -->
-    <script async src="https://www.googletagmanager.com/gtag/js?id={{analyticsId}}"></script>
-  {{end}}
-{{end}}
-
 {{block googleScripts(tagManagerId=config("google_tag_manager_id"), analyticsId=config("google_analytics_id"))}}
   <!-- Google integration scripts -->
   <script>
     function loadGoogleTagManager() {
-      (function(w,d,s,l,i){w[l]=w[l]||[];w[l].push({'gtm.start':
-      new Date().getTime(),event:'gtm.js'});var f=d.getElementsByTagName(s)[0],
-      j=d.createElement(s),dl=l!='dataLayer'?'&l='+l:'';j.async=true;j.src=
-      'https://www.googletagmanager.com/gtm.js?id='+i+dl;f.parentNode.insertBefore(j,f);
-      })(window,document,'script','dataLayer','{{ tagManagerId }}')
+      (
+        function(w,d,s,l,i){
+          w[l]=w[l]||[];w[l].push({'gtm.start':new Date().getTime(),event:'gtm.js'});
+          var f=d.getElementsByTagName(s)[0],
+          j=d.createElement(s),
+          dl=l!='dataLayer'?'&l='+l:'';j.async=true;j.src='https://www.googletagmanager.com/gtm.js?id='+i+dl;f.parentNode.insertBefore(j,f);
+          }
+      )
+      (window,document,'script','dataLayer','{{ tagManagerId }}')
     }
 
     <!-- GA4 code snippet if G tracking code is provided, else use legacy UA code snippet -->
     {{if hasPrefix(analyticsId,"G")}}
       function loadGoogleAnalytics() {
+        (
+          function(d, googleId){
+          var f = d.getElementsByTagName('script')[0];
+          var s = d.createElement('script');
+          s.src = 'https://www.googletagmanager.com/gtag/js?id=' + googleId;
+          s.async = true;
+          f.parentNode.insertBefore(s,f);
+          }
+          (document,'{{analyticsId}}')
+        );
+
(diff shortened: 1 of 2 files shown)


site/templates/application/google.jet
```diff
@@ -1,24 +1,32 @@
...
 {{block googleScripts(tagManagerId=config("google_tag_manager_id"), analyticsId=config("google_analytics_id"))}}
   <!-- Google integration scripts -->
   <script>
     function loadGoogleTagManager() {
-      (function(w,d,s,l,i){w[l]=w[l]||[];w[l].push({'gtm.start':
-      new Date().getTime(),event:'gtm.js'});var f=d.getElementsByTagName(s)[0],
-      j=d.createElement(s),dl=l!='dataLayer'?'&l='+l:'';j.async=true;j.src=
-      'https://www.googletagmanager.com/gtm.js?id='+i+dl;f.parentNode.insertBefore(j,f);
-      })(window,document,'script','dataLayer','{{ tagManagerId }}')
+      (
```
l0ud0gg, line 5: ah i de-uglified this to make sense of it. do i need to revert it?

site/templates/application/google.jet
```diff
@@ -1,24 +1,32 @@
...
+          dl=l!='dataLayer'?'&l='+l:'';j.async=true;j.src='https://www.googletagmanager.com/gtm.js?id='+i+dl;f.parentNode.insertBefore(j,f);
+          }
+      )
+      (window,document,'script','dataLayer','{{ tagManagerId }}')
     }
 
     <!-- GA4 code snippet if G tracking code is provided, else use legacy UA code snippet -->
     {{if hasPrefix(analyticsId,"G")}}
       function loadGoogleAnalytics() {
+        (
```
l0ud0gg, line 19: should i uglify this bit also?

site/templates/application/google.jet
stajs: Yeah, revert back to the ugly since it's what we've copied from Googles snippets.

site/templates/application/google.jet
stajs: Yeah I think we should because it's in every page, might as well save bytes.

stajs (changes requested):
U-G-L-Y, you ain't got no alibi!

PeculiarGoat (changes requested):
What Shannon said!

PeculiarGoat (approved):

site/templates/application/google.jet
```diff
@@ -1,24 +1,19 @@
...
-      (function(w,d,s,l,i){w[l]=w[l]||[];w[l].push({'gtm.start':
-      new Date().getTime(),event:'gtm.js'});var f=d.getElementsByTagName(s)[0],
-      j=d.createElement(s),dl=l!='dataLayer'?'&l='+l:'';j.async=true;j.src=
-      'https://www.googletagmanager.com/gtm.js?id='+i+dl;f.parentNode.insertBefore(j,f);
-      })(window,document,'script','dataLayer','{{ tagManagerId }}')
-    }
+      (function(w,d,s,l,i){w[l]=w[l]||[];w[l].push({'gtm.start':new Date().getTime(),event:'gtm.js'});
+      var f=d.getElementsByTagName(s)[0],j=d.createElement(s),dl=l!='dataLayer'?'&l='+l:'';
+      j.async=true;j.src='https://www.googletagmanager.com/gtm.js?id='+i+dl;f.parentNode.insertBefore(j,f);
+      })(window,document,'script','dataLayer','{{ tagManagerId }}')}
```
stajs, line 8: I know this has already merged, but that last `}` should be a new line since it's not part of the IIFE, it's the closing brace for the `function loadGoogleTagManager() {`

stajs: A part of me also wants the semicolon even though it's not strictly necessary since it's a single statement (the IIFE), e.g.:

```
  ... '{{ tagManagerId }}');
}
```

site/templates/application/google.jet
```diff
@@ -1,24 +1,19 @@
...
 
     <!-- GA4 code snippet if G tracking code is provided, else use legacy UA code snippet -->
     {{if hasPrefix(analyticsId,"G")}}
       function loadGoogleAnalytics() {
+        (function(d, googleId){var f = d.getElementsByTagName('script')[0];var s = d.createElement('script');
+        s.src = 'https://www.googletagmanager.com/gtag/js?id=' + googleId;s.async = true;f.parentNode.insertBefore(s,f);
+        }(document,'{{analyticsId}}'));
```
stajs, line 15: You could have been more aggressive and reduced spaces/var name

```js
        (function(d,id){var f=d.getElementsByTagName('script')[0];var s=d.createElement('script');
        s.src='https://www.googletagmanager.com/gtag/js?id='+id;s.async=true;f.parentNode.insertBefore(s,f);
        }(document,'{{analyticsId}}'));
```

stajs (commented):
Yeah, looks fine, but maybe a couple of things could also tidy later.

PR #316, opened 2022-04-11
Implement the designs - two column infos on tv/film detail page

matthewthomsonnz:
- The sole purpose of this PR is to turn the tv/film detail page info items into two columns as per designs
- Just utilises bootstrap `.row` and `.col-lg-6` to achive this
- [AB#6846](https://dev.azure.com/S72/fefacba9-96b6-4af2-a53d-050ed453e13a/_workitems/edit/6846)

image


brynyShift72 (commented):
Probably not very likely to happen (or a big deal), but do want the headings in be inline with each other like currently are? It seems that if we include a large cast for example, we are leaving a huge gap between one columns headers. 
image

brynyShift72:
Well that image was bigger than expected..

matthewthomsonnz:
> Probably not very likely to happen (or a big deal), but do want the headings in be inline with each other like currently are? It seems that if we include a large cast for example, we are leaving a huge gap between one columns headers. 

@brynyShift72 Designs indicate headings inline, and looking through a couple of clients 7 was the most cast I could find.

Here's a movie with the most I could think of (the expendables with 11 cast) which still looks looks acceptable
image

brynyShift72 (approved):
LGTM, might want to wait for the css gods @curtisadams28 @PeculiarGoat before merging.
In before design team wants a ellipses/ expandable sections if there are too many elements

site/templates/film/item.jet
```diff
@@ -70,28 +70,27 @@
...
-              {{end}}
-                <p>
-                  {{range index, member := film.Cast}}
-                    {{ path := "/search.html?q=" + member.Name }}
-                    <a href="{{ routeToPath(path) }}">{{member.Name}}</a>{{if index < len(film.Cast) - 1}}, {{end}}
-                  {{end}}
-                </p>
-              </div>
-              <div class="meta-detail-crew d-flex">
+                <div class="mb-2 row">
```
PeculiarGoat, line 73: Don't think that div needs to be indented there!

matthewthomsonnz: Ah I see that div just before it closes on the same line. Thanks - have unindented.

PeculiarGoat (changes requested):
I've had a think about this and decided Bootstrap isn't cool.  Coming from a perspective of customising a site (whether done by us or by an external dev), overriding Bootstrap CSS is a massive pain compared to overriding our own core-template CSS (it's easier to find and _usually_ much easier to comprehend).

IMO this would be better done with flex-wrap and with all of those Bootstrap classes removed (maybe stop using them elsewhere too :P).  Thoughts @stajs ?

matthewthomsonnz:
> I've had a think about this and decided Bootstrap isn't cool. Coming from a perspective of customising a site (whether done by us or by an external dev), overriding Bootstrap CSS is a massive pain compared to overriding our own core-template CSS (it's easier to find and _usually_ much easier to comprehend).
> 
> IMO this would be better done with flex-wrap and with all of those Bootstrap classes removed (maybe stop using them elsewhere too :P). Thoughts @stajs ?



> I've had a think about this and decided Bootstrap isn't cool. Coming from a perspective of customising a site (whether done by us or by an external dev), overriding Bootstrap CSS is a massive pain compared to overriding our own core-template CSS (it's easier to find and _usually_ much easier to comprehend).
> 
> IMO this would be better done with flex-wrap and with all of those Bootstrap classes removed (maybe stop using them elsewhere too :P). Thoughts @stajs ?

@PeculiarGoat Makes sense, just need a strategy for margin -15px  row and 15px on column. Would you suggest CSS grid instead now we've dropped IE?

IMO often not a fan of frivolous custom css like that, especially for one-off patterns. Can get unwieldy and hard for newcomers.

PeculiarGoat:
@matthewthomsonnz Yeah CSS grid can probably do it good.  In general I would agree with your comment about frivolous custom CSS... but when doing custom site work with shift72-template, it's always easier if the CSS is all nicely contained within core-template... maybe I'm the only one that feels that way? lol

stajs:
In terms of philosophical direction, I don't think Bootstrap gives us enough benefit to warrant the current layers of shit surrounding it, and with us dropping IE support I think long term we should be questioning if we need it still. But currently it's tentacles are pretty deep so I don't have a major objection either way.

PeculiarGoat:
It would be nice if we can transition away from it.

matthewthomsonnz:
- Removed bootstrap col/row. Added CSS grid equivalent.

It really is hard to think of way out of bootstrap, an ideas doc and  a spike could be the way

site/styles/_meta-detail.scss
```diff
@@ -590,3 +576,15 @@
     font-size: 12px;
   }
 }
+
+s72-element-switcher {
+  @extend .mb-5;
```
PeculiarGoat, line 581: Can you replace this with `margin-bottom`... `mb-5` is a one liner anyway so we don't gain anything by obfuscating.

site/styles/_meta-detail.scss
```diff
@@ -590,3 +576,15 @@
...
+}
+
+.meta-detail-infos {
+  @include media-breakpoint-up(lg) {
+    display: grid;
+    grid-template-columns: repeat(2, 1fr);
+    gap: 15px;
+  }
+}
```
PeculiarGoat, line 590: Probably nicer to consistently use grid across breakpoints:
```suggestion
.meta-detail-infos {
  display: grid;
  grid-template-columns: repeat(1, 1fr);
  gap: 20px;

  @include media-breakpoint-up(lg) {
    grid-template-columns: repeat(2, 1fr);
  }
}
```

PeculiarGoat: Also... if you change the `gap` to `20px` and remove `padding-bottom` from line 279, it matches the designs a bit better.

site/templates/film/item.jet
```diff
@@ -70,26 +70,25 @@
             {{end}}
             <div class="{{ isEnabled("element_switcher_enabled") ? "s72-hide" : "" }}"> {* hide class prevents FOUC *}
               <div class="meta-detail-synopsis">{{film.Overview | raw}}</div>
-              <div class="meta-detail-cast">
+              <div class="meta-detail-infos">
               {{if len(film.Cast) > 0 }}
```
PeculiarGoat, line 74: This block needs an indent!

site/templates/tv/detail.jet
```diff
@@ -47,20 +47,20 @@
...
-          {{end}}
-            <p>
-              {{range index, member := tvseason.Cast}}
-                {{ path := "/search.html?q=" + member.Name }}
-                <a href="{{ routeToPath(path) }}">{{member.Name}}</a>{{if index < len(tvseason.Cast) - 1}}, {{end}}
-              {{end}}
-            </p>
-          </div>
-          <div class="meta-detail-crew d-flex">
+          <div class="meta-detail-infos">
```
PeculiarGoat, line 50: Can you please add the element switcher to this file too and move the cast `if` statement outside the cast div to match the film detail page.

PeculiarGoat (changes requested):
Grid is much nicer!  Just a few little things to look at:

site/styles/_meta-detail.scss
```diff
@@ -590,3 +576,15 @@
     font-size: 12px;
   }
 }
+
+s72-element-switcher {
+  @extend .mb-5;
+}
+
+.meta-detail-infos {
```
stajs, line 584: Nit: I think should be singular. It doesn't matter the class is used multiple times.

```suggestion
.meta-detail-info {
```

stajs (approved):
As long as @PeculiarGoat's comments are addressed, LGTM.

site/styles/_meta-detail.scss
matthewthomsonnz: - replaced

site/styles/_meta-detail.scss
matthewthomsonnz: - suggestion committed and padding-bottom removed

site/templates/film/item.jet
matthewthomsonnz: - indent added

site/styles/_meta-detail.scss
matthewthomsonnz: - changed to singular

site/templates/tv/detail.jet
matthewthomsonnz: - element switcher functionality added to tv detail
- if statement matched to film detail (outside cast)

PeculiarGoat (approved):
Looks good!  Nice work!

PR #324, opened 2022-04-26
Fix: Handle multiple sentences in script

josejibin:
Seems like Transalate API return response in array if there are more
than one sentence.

API sample:
https://translate.googleapis.com/translate_a/single?client=gtx&sl=en&tl=fr&dt=t&q=First sentance. Second sentance
vs
https://translate.googleapis.com/translate_a/single?client=gtx&sl=en&tl=fr&dt=t&q=First sentance Second sentance


stajs (approved):
Good find. This looks to be the indexing:

image

PR #337, opened 2022-05-09
AB#7964: fix wordings on bundle

josejibin:
[
ADO Card](https://dev.azure.com/S72/SHIFT72/_boards/board/t/Web%20team/Stories/?workitem=7964)

Currently, bundles only show film count. This PR will fix that.

* Removed `bundle_items_all_films`, `bundle_items_all_seasons` translation keys.
* Added `film_count`, `season_count` translation keys.
* Edited `bundle_items_mixed` to use `film_count`, `season_count` for plurals.

This PR depends on the [kibble PR](https://github.com/shift72/kibble/pull/126) which adds the `getCount` function to items. Relish also depends on this PR

### DIFF
--- a/site/ar_LB.all.json
+++ b/site/ar_LB.all.json
@@ -794,12 +794,6 @@
   "meta_detail_bundle_items_title": {
     "other": "المحتوى المتضمن في هذه الحزمة"
   },
-  "bundle_items_all_films": {
-    "other": "{{.Count}} أفلام"
-  },
-  "bundle_items_all_seasons": {
-    "other": "{{.Count}} مواسم"
-  },
   "bundle_items_mixed": {
     "other": "{{.FilmCount}} و {{.SeasonCount}}"
   },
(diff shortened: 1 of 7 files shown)


PeculiarGoat:
IMO we shouldn't be using compound translations because they just create problems later when things get changed.  I wonder if we can have a chat with the design team and propose that we have one translation for "_n_ items" instead of the three in this PR (film, season, mixed).  Then that `GetCount` function in the kibble PR could be made intuitive by having it return an int instead of a map.

Also... translations that use a `{{.Count}}` variable must have _all_ pluralities as defined for each language [here](https://github.com/nicksnyder/go-i18n/blob/0dc1626d56435e9d605a29875701721c54bc9bbd/i18n/language/codegen/plurals.xml)... I see you have removed some in this PR... they will need to be put back!

matthewthomsonnz:
I feel like the mixed state is complicateds things

matthewthomsonnz:
I feel like the mixed state complicates things, it should just be simple two translations + getCount resulting in: `n film(s), n tv season(s)` 

Ps sorry accidentally pressed some buttons in this PR

PR #342, opened 2022-05-12
Ab#8003 submit button is hidden on pin modal

matthewthomsonnz:
[AB#8003](https://dev.azure.com/S72/SHIFT72/_boards/board/t/Web%20team/Stories/?workitem=8003)

What happened was:
- We changed the sitewide primary button text color to black
- However there were some buttons with a black background, making the text invisible

I had a good look and think I've covered them all in this fix.


l0ud0gg (approved):
LGTM. We prolly need to be careful on upgrades to make sure their buttons / primary colour dont clash.

PR #345, opened 2022-05-16
Add unqiue class to tagline

brynyShift72:
[AB#8005](https://dev.azure.com/S72/fefacba9-96b6-4af2-a53d-050ed453e13a/_workitems/edit/8005) Stemmed from Sakka request to hide runtime on front end and keep other taglines. Would be nice to add to core template.
All meta items were previously the same class so hiding an individual one required either
1. Removing metadata from /admin page (requires configuration to skip publishing validation)
2. Removing specific logic from `tagline.jet`

Notes: I think some styles are being pulled through relish with an `!important` property eg, `display: inline !important;`, meaning that making these items disappear would require a `display: none !important;` (Maybe there is a better way to do this)

Would also be open to better class names

### DIFF
--- a/site/templates/items/tagline.jet
+++ b/site/templates/items/tagline.jet
@@ -34,30 +34,30 @@
     {{end}}
 
     {{if isset(.Runtime)}}
-      {{yield taglineItem() content}}
+      {{yield taglineItem(class="meta-item-tagline-runtime") content}}
         {{if .Runtime > 60}}
           {{.Runtime.Hours()}}{{i18n("runtime_hours")}} {{.Runtime.Minutes()}}{{i18n("runtime_minutes")}}
         {{else}}
           {{.Runtime}}{{i18n("runtime_minutes")}}
         {{end}}
       {{end}}
-      {{yield taglineDivider(dividers=dividers)}}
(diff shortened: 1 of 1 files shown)


PeculiarGoat:
@brynyShift72 this would be more readable if you ditched the `meta-item-tagline-` from those new classes.  You can probably live without the `-divider` suffix too.  Thoughts?

l0ud0gg (approved):
good call. looks good to me.

stajs:
I like the idea and looking at the file has prompted some refactor thoughts, so I took a run at it also but including @PeculiarGoat's points about class name brevity.

curtisadams28:
I wonder if hiding the parts of the tagline would be better done with uber configs?

stajs:
Why not both? If uber, changes still have to come through us and wait until we build out admin UI. This will be available once self-service CSS is out.

curtisadams28:
Good point.

curtisadams28 (changes requested):
I'm just going to leave this here so it doesn't show up in PR shaming. Guessing we're going to use Shannons [PR](https://github.com/shift72/core-template/pull/347)?

PeculiarGoat:
@brynyShift72 bump!

PR #347, opened 2022-05-18
Tagline refactor

stajs:
Basically a rip of @brynyShift72's idea from #345 and a refactor.


site/templates/items/tagline.jet
```diff
@@ -9,85 +13,56 @@
...
         {{itemGenres}}
       {{end}}
-      {{yield taglineDivider(dividers=dividers)}}
-      {{dividers = dividers - 1}}
+      {{showDivider = true}}
     {{end}}
 
-    {{if isset(.ReleaseDate) && .ReleaseDate.Year() > 1}}
-      {{yield taglineItem() content}}
+    {{if isset(.ReleaseDate) && .ReleaseDate.Year() > 1888}}
```
stajs, line 52: `1888` = first film ever 📽️

site/templates/items/tagline.jet
```diff
@@ -9,85 +13,56 @@
...
-  {{if dividers > 0}}
-    <span class="meta-item-tagline-divider">•</span>
-  {{end}}
+{{block item(class="missing", showDivider=showDivider)}}
+  <span class="{{class}}">
+    {{if showDivider}}
+      <span class="divider">•</span>
+    {{end}}
+    <span class="item">{{yield content}}</span>
+  </span>
```
stajs, line 67: This changes the DOM output from:

```html
<item1><divider><item2><divider><item3>
```

... to:

```html
<wrapper><item1></wrapper>
<wrapper><divider><item2></wrapper>
<wrapper><divider><item3></wrapper>
```

stajs: image

stajs: Note the whitespace around `item`'s - that's pre-existing. We can remove whitespace in the jet so it doesn't appear in the final output but it makes the code less readable so left as is. 

There is [whitespace trimming](https://github.com/CloudyKit/jet/blob/master/docs/syntax.md#whitespace-trimming) that could potentially make use of but doesn't seem to available in the currently used version - something for the v2 Jet upgrade.

stajs: I think the DOM structure gives good CSS targets.

PeculiarGoat: It gives good CSS targets for Bryny's goal (hide an item and a divider with one selector), but isn't this a step back from the flat DOM structure for margins between everything?

To put the margins back in, you would need to do something like:
```
.meta-item-tagline {
  > * {
    &:not(:last-child, :only-child, s72-classification-label) {
      margin-right: map-get($spacers, 1); // Currently 0.25rem
    }

    .divider {
      margin-right: map-get($spacers, 1); // Currently 0.25rem
    }
  }
}
```
Which is gross... though only a bit grosser than the existing CSS.

But if we keep the existing flat DOM structure, we can refactor the CSS with `gap` and it will be much cleaner and easier to work on in the future.

site/templates/items/tagline.jet
```diff
@@ -9,85 +13,56 @@
...
-    {{if isset(.ReleaseDate) && .ReleaseDate.Year() > 1}}
-      {{dividers = dividers + 1}}
-    {{end}}
-    {*{ There should be one less dividers than tagline items. }*}
-    {{dividers = dividers - 1}}
-    
-    {{if classification}}
-      <s72-classification-label data-slug="{{.Slug}}" data-layout="tooltip"></s72-classification-label>
-    {{end}}
+    {{showDivider := false}}
```
PeculiarGoat, line 16: This is a much better way!

site/templates/items/tagline.jet
PeculiarGoat: Heh too smart for me lol.  👍  for margins on dividers only!

l0ud0gg:
i like the idea of cleaning up the dom and having more functional classes to utilise. Do the existing css styles need to be refactored to handle this new setup? Also we might have to be careful if we refactor this jet and styles with clients that have custom tagline templates. upgrades might break them if we arent careful (some older clients have custom tagline.jet to add / remove fields in the tagline)

l0ud0gg (commented):
as commented

curtisadams28 (approved):

brynyShift72 (approved):

site/templates/items/tagline.jet
brynyShift72: No big deal either way but wondering if we want this to be any number for a clients "artistic" reasons

PeculiarGoat:
@stajs bump!

stajs:
@curtisadams28 Can you just eyeball the changelog for me please.

curtisadams28 (approved):

PR #348, opened 2022-05-19
Fix issue caused by linter moving media query

matthewthomsonnz:
**TLDR: stylelint moved some code, broke bonus/episodes style. Move back and add lint ignore rule.**

- I forgot to add an ignore to stylelint, so when someone else ran it, it moved a media query ([commit](
https://github.com/shift72/core-template/commit/59ddbd13#diff-8621f91c695f61e8fd0f3a81890fd5251371d9eacabd842ed490500ad5403e45R9-R29
)) to above the code it was targeting, making the media query overridden, resulting in layout issues. This PR moves the code back and adds an ignore.

Note: I had to just use /* stylelint-disable */ sorry - couldn't find the exact stylelint rule that moves the code block.

### DIFF
--- a/site/styles/_meta-sub-item.scss
+++ b/site/styles/_meta-sub-item.scss
@@ -6,27 +6,6 @@
   margin-bottom: 15px;
   padding-bottom: 15px;
 
-  @include media-breakpoint-up(sm) {
-    grid-template-columns: 175px minmax(100px, 1fr);
-    grid-template-rows: auto auto minmax(0, 1fr);
-
-    .synopsis {
-      grid-column: auto;
-      grid-row: 2;
-      margin-bottom: 10px;
-    }
-    .poster-container {
-      grid-row: 1 / 4;
-    }
-    .runtime {
-      margin-bottom: 0;
-      margin-left: 0;
-    }
-    h3 {
-      margin-bottom: 0;
-    }
-  }
-
   h3 {
     color: var(--body-color);
     font-size: 16px;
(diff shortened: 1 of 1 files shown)


stajs (approved):
Crazy!

@PeculiarGoat @l0ud0gg @brynyShift72 @curtisadams28 We should investigate the lint rule that moves these and turn it off.

PeculiarGoat:
I'm of the opinion that each selector should have it's own media query if it needs one rather than a media query wrapping the selectors and becoming order dependant.  Wrapping things in a media query like this makes it harder to maintain... for example, the next person to change the `.synopsis` styles will have to do it in two places instead of one... assuming they even realise that `.synopsis` styles are defined in two places.  I think if the CSS breaks after being linted, that should be taken as a hint that the approach is off.

stajs:
@PeculiarGoat Yeah, I agree the selector wrapping the media query is probably preferred most of the time. But given we've got existing code I don't understand why the linter would decide to change the order, from:

```scss
.sub-item {
  h3 { ... }
  ...
  .synopsis { ... }
  @include media-breakpoint-up(sm) { .synopsis ... }
}
```

... to:

```scss
.sub-item {
  @include media-breakpoint-up(sm) { .synopsis ... }
  h3 { ... }
  ...
  .synopsis { ... }
}
```

It seems a dangerous change that could break styles in non-obvious ways. If it linted something you weren't working on and couldn't spot the effects of the change, I suspect most people would just trust the linter and commit the change and that's probably what happened for this issue.

PeculiarGoat:
@stajs it's because the linter wants attributes first, then media queries, then children.  The expectation is that media queries are going to modify attributes on the owning selector that you have defined directly above so it's all easier to understand.  You wouldn't want to have to scroll up and down repeatedly to mentally parse how the media query changes the default attributes.

PR #353, opened 2022-05-23
bump node version in CI

PeculiarGoat:
Bump node version to 16 in `node.js.yml` to stop `npm ci` failing in github actions.


stajs (approved):
Good, but an even better option is probably to also match minor and patch from `.nvmrc` too.

PR #354, opened 2022-05-23
AB#8128 Pass explicit paths into share modals

additiverich:
ADO card: ☑️ [AB#8128](https://dev.azure.com/S72/SHIFT72/_boards/board/t/Web%20team/Stories/?workitem=8128)

- [❌] Has this been discussed with Delivery Team?

## Description of work
- Currently the share modal "shares" the current URL (i.e. `window.location.href`)
- This means clicking the share modal from the homepage carousel shares the homepage URL to Facebook, Twitter etc
- Support for an explicit attribute "path" in the ShareModal component has been introduced into Relish (under review)
- Template function `socialMediaButtons` had support for passing a path into the social media buttons but it wasn't used for anything
- Pass this path into the share modal
- If this is empty or missing it falls back to `window.location.href` in the component
- Update files film/item.jet, bundle/extras.jet, tv/detail.jet and carousel/item/extras.jet to include URLs (including domain name) when calling `socialMediaButtons`
- Possibly net positive of explicit URLs is removing query string parameters from URLs

(none)

(nothing new)

## Related PRs 
https://github.com/indiereign/shift72-web/pull/307

- Relish - as above

### Affected Clients
 - I don't think the new carousel is rolled out yet

## Checklist
- [✅] CI tests are passing Github actions (inc. linting)
- [✅] Moved ADO card to Dev/done, checked link to Github, tagged "Review"
- [✅] Updated changelog (if applicable)
- [✅] I promise to document any new feature toggles/configurations in the appropriate documentation

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ No changes from `alpha-0`.
 ### Added
 - Keyboard controls for the carousel.
 - Pull request template file.
+- `path` parameter in share modals
 
 ### Changed
 - Reworked the carousel UI.
--- a/site/templates/bundle/extras.jet
+++ b/site/templates/bundle/extras.jet
@@ -8,6 +8,6 @@
 
     <s72-userwishlist-button data-slug="{{slug}}" class="btn-wishlist"></s72-userwishlist-button>
 
-    {{yield socialMediaButtons(path=currentUrlPath)}}
+    {{yield socialMediaButtons(path=site.SiteConfig.SiteURL + routeToSlug(slug))}}
   </div>
 {{end}}
--- a/site/templates/bundle/extras.jet
+++ b/site/templates/bundle/extras.jet
@@ -8,6 +8,6 @@
 
     <s72-userwishlist-button data-slug="{{slug}}" class="btn-wishlist"></s72-userwishlist-button>
 
-    {{yield socialMediaButtons(path=site.SiteConfig.SiteURL + routeToSlug(slug))}}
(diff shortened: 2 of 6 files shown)


PeculiarGoat (commented):
This looks fine.  But IMO the url + path concat should be handled in the component rather than in jets.  Is there a specific reason to have it in the jets?

additiverich:
> This looks fine. But IMO the url + path concat should be handled in the component rather than in jets. Is there a specific reason to have it in the jets?

I think it should honour the URL which is passed into it, rather than assuming it's a page on "this" site and needs the domain prepended to it

PeculiarGoat:
Fair enough.  In that case, should we change `path` to `url` or something?  Maybe too nitpicky?

additiverich:
Could do; "path" was already being used. Relish seems to be split evenly between the two.

PeculiarGoat (approved):
Happy with this 👍 
This is going to be blocked on the Relish PR isn't it?

additiverich:
@PeculiarGoat It'll behave as it currently does until Relish is updated

CHANGELOG.md
```diff
@@ -9,6 +9,7 @@ No changes from `alpha-0`.
 ### Added
 - Keyboard controls for the carousel.
 - Pull request template file.
+- `path` parameter in share modals
```
stajs, line 12: Just a small typo, if we stick with a URL, we'll have to update in a future PR. However I'm also thinking we should just pass the slug through to the block and it can construct the URL there.

PR #355, opened 2022-05-24
Carousel right aligned backgrounds

curtisadams28:
ADO card: ☑️ [AB#8123](https://dev.azure.com/S72/SHIFT72/_workitems/edit/8123)

## Description of work
Allows carousel images to be right aligned with a left gradient. This can be used as a failsafe if clients want their images to continue being right aligned after we upgrade them.

This work is currently deployed to [staging-test](https://staging-test.shift72.com).

- `Uber Admin > Users > Configuarations > carousel_image_right_alignment`. Set to `true` (public)

## Checklist
- [ ] Design review

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 ### Added
 - Keyboard controls for the carousel.
 - Pull request template file.
+- Carousel background can now be right-aligned or centered.
 
 ### Changed
 - Reworked the carousel UI.
--- a/kibble.json
+++ b/kibble.json
@@ -1,7 +1,7 @@
 {
   "name": "core-template",
   "version": "0.0.1",
-  "siteUrl": "https://staging-test.shift72.com",
+  "siteUrl": "https://abccinemas.screenplus.co",
   "builderVersion": "0.16.6",
   "defaultLanguage": "en",
   "languages": {
--- a/kibble.json
+++ b/kibble.json
@@ -1,7 +1,7 @@
 {
   "name": "core-template",
   "version": "0.0.1",
-  "siteUrl": "https://abccinemas.screenplus.co",
+  "siteUrl": "https://staging-test.shift72.com",
   "builderVersion": "0.16.6",
   "defaultLanguage": "en",
   "languages": {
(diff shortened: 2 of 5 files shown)


stajs (changes requested):
A config doesn't feel right for CSS positioning. I was thinking a CSS var e.g. `--carousel-image-object-position: center` that we could override with custom CSS to set the [object-position](https://developer.mozilla.org/en-US/docs/Web/CSS/object-position). I thought that wouldn't need any jet mess and also means you could have a more finely-controlled positioning with percentage/pixels if needed.

curtisadams28:
@stajs Right. The the problem with object-position is that the image will still be centered once the window is larger that the image dimensions. It only works where it can cut off the image more on one side than the other. It does nothing on 1920x1080 displays. Is it still worth having?

With my solution, if you want the image properly right aligned, then you need to add the gradient back in for when the window size is greater than the image width. I coud use sass instead of jet but I don't think it would be any less hax.

Maybe it would be better to use custom css if the client wants images right aligned?

l0ud0gg:
> @stajs Right. The the problem with object-position is that the image will still be centered once the window is larger that the image dimensions. It only works where it can cut off the image more on one side than the other. It does nothing on 1920x1080 displays. Is it still worth having?
> 
> With my solution, if you want the image properly right aligned, then you need to add the gradient back in for when the window size is greater than the image width. I coud use sass instead of jet but I don't think it would be any less hax.
> 
> Maybe it would be better to use custom css if the client wants images right aligned?

does make things a little complicated supporting both imho. I think the goal of the right alignment is to make sure the focus of the carousel image is maintained at lower resolutions? Would a simpler solution be to have everyone using full width at desktop res and implementing the right align just for tablet / mobile res?

curtisadams28:
Ok. I'm going to close this PR and create a new branch.

stajs:
Don't forget to delete the branch if you don't need!

At 1920 viewport the image being centred horizontally is true and fine - you've got the full width to see the entire picture! However having the option to adjust the vertical postion could be useful for some, e.g. instead of a 1600 image being cropped top and bottom equally, could choose to crop just the top, just the bottom, or some other pixel position.

At smaller viewports the horizontal position is probably more important for clients, especially those with 1600 images.

PR #359, opened 2022-05-26
Headings spacing 8089

matthewthomsonnz:
ADO card (headings): ☑️ [AB#8089](https://dev.azure.com/S72/SHIFT72/_workitems/edit/8089)

Designs: 
- 🖥 [typography designs section](https://zpl.io/amxOMBr)
- 🖥 [Desktop/tablet](https://app.zeplin.io/project/5fdaa165fbacf584a2ce4f8b/screen/61231fda5d02ef1252e998e7)  (font stylesheet)
- 📱 [Mobile](https://app.zeplin.io/project/5fdaa165fbacf584a2ce4f8b/screen/61232090efca7116d70899fe)  (font stylesheet)

## Description of work
The purpose of this PR is:
- Uniform font styles throughout the template
- Remove many scoped css styles
- Tweak spacing between elements, as well as implement meta-detail page top spacing
- Compliant element structure for [WCAG](https://www.w3.org/WAI/tutorials/page-structure/headings/) and [SEO](https://developers.google.com/style/headings#heading-and-title-format) (mainly just valid h1, h2, h3 etc structure) 
- and instead adopt strategy of e.g. `<h1 class="h2-style">`

Note:
- You'll notice css rules do not match the specifications in the font stylesheet. This is due to discrepancies between design applications and web browsers. Instead opting for visual accuracy by matching the fonts using a transparent browser window over the designs.
- I've also taken into account difference in how font weights are rendered between operating systems, and have chosen one weight lighter in a couple of cases so it looks closer to the designs
- Please ignore the following for this PR:
  - award categories text, sponsor logo text, all buttons
  - 'label' 'caption 2' 'button' (on design's typography styleguide)


site/templates/tv/detail.jet
```diff
@@ -80,7 +80,7 @@
 
           {{if len(tvseason.Episodes) > 0}}
             <div class="meta-detail-episodes-content">
-              <h2>{{i18n("meta_detail_episodes_title")}}</h2>
+              <h2 class="h3-style">{{i18n("meta_detail_episodes_title")}}</h2>
```
smitsanghvi21, line 83: I'm being bit fussy but the` <h1>` with class name as h2` <h2>` with class name as h3 looks bit confusing

site/styles/_typography.scss
```diff
@@ -0,0 +1,116 @@
...
+  letter-spacing: 0.5px;
+  line-height: 23px;
+}
+
+h1 {
+  color: var(--body-color);
+  font-size: 35px;
+  font-weight: $font-weight-normal;
+  letter-spacing: 0.8px;
+  margin-bottom: 8px;
```
PeculiarGoat, line 21: I don't think structural CSS (in this case `margin`) should be in a file dedicated to typography.

matthewthomsonnz:
> Welcome to my TED talk.
> 
> I'm feeling a bit skeptical about this approach. Feel free to argue though.
> 
> * I'm getting a sort of OOCSS vibe from the new typography classes and I feel it doesn't fit well with the existing CSS approach (Sass nesting apocalypse).

@PeculiarGoat  I think you'll understand and (I'm hoping agree) if I explain the bigger picture:

**Problem:** We need consistency in spacing and typography to move towards something resembling [atomic design](https://bradfrost.com/blog/post/atomic-web-design/) (also potentially implementing [storybook](https://storybook.js.org/ ) but forget that for now).

**Proposed Solution:** Minimal global/OOCSS styles for the 'atoms' (in this PR). Which compliment, (not replace, but provide a solid foundation for), your existing component scoped style strategy (component being say '.carousel', or '.app-banner') for Molecules/Organisms/Templates/Pages (as you already have).

> You'll notice that our classes get their names based on what the element is for example `app-badge-title` is an `h2` element which contains the app badge title.  I see you've added a `paragraph-style` class on that element and it makes me wonder how it can be a paragraph and a title at the same time.  And it looks like that is the only usage of `paragraph-style`... why not select with the existing specific class `app-badge-title`?  Another example is the copyright statement element.  It's a `p` so I would assume it would have the `paragraph-style` class... but instead it has `body-2-style`... yet it isn't a `body` element and the `2` part doesn't seem to be relevant?  Also, I see an `h2` element with an `h3-style` class?!

 I see how this could be confusing. What is the solution? Perhaps we use `@extend` then with a naming convention that references design's typography style guide. e.g. type-style-1, heading-style-c 


> * This could generate even more pain around the entanglement of core-template and Relish releases.  If I spot an element inside a Relish component and decide it needs one of these new typography classes e.g. I see a `p` and I want to give it the `paragraph-style` class, I now have go through the whole PR/test/release process with Relish.  That seems pretty overkill when you consider that the current process to do the same thing would be to simply target the element using it's existing specific class in the core-template CSS... no need to touch Relish at all.

This is an existing issue that is inherent with relish regardless of this PR. You'll notice I've used extend as we've always done for these cases. e.g. on [s72-availability-label](https://github.com/shift72/core-template/pull/359/files#diff-1690b8cb885e67b64253816006cb361c5af104c3073fe62ebfb567c91815dcd5R5), [.s72-form-check-label and .s72-form-control.](https://github.com/shift72/core-template/pull/359/files#diff-420814d912bccbbb5edf484a806569b12b800bd343f5f40e73f4d2eda710593cR110-R113)

> 
> My point is... I think diverging from our existing CSS strategy, even though it's not perfect, is going to make things even more confusing. A better approach would be to do this entirely in the Sass files with no class changes in the jets e.g:
> 
> ```
> .app-badge-title {
>   @extend .paragraph-style;
> }
> ```

I will switch to this approach. Inline helper classes is me pushing my opinionated strategies. Here is a screenshot of my main frustration with `@extend` though:
Screenshot from 2022-05-30 12-36-23



> 
> You could also cut down the number of lines in the typography file quite a bit by adding mixin since all of those classes have basically the same properties.

Mixin added

site/templates/tv/detail.jet
stajs: Yeah, agree, this is a no from me on this.

site/styles/_mixins.scss
```diff
@@ -77,7 +77,6 @@
   &:hover,
   &:focus {
     color: var(--link-color);
-    text-decoration: none;
```
stajs, line 80: This seems like a rather small change, but why? This doesn't appear to be the main point of this PR.

stajs (changes requested):
This seems a huge change in how the CSS is fundamentally structured to drop in a PR called "headings spacing". I don't think this is the approach we should be taking.

The positives:
- Agree need to make typography consistent.
- Agree need to sort out heading hierarchy.
- Agree `@extends` has lead to some awful CSS.
- @PeculiarGoat has previously advocated for atomic design, so you have an ally there.

The negatives:
- This is not the approach to switch to atomic design - that is a discussion that needs to happen.
- `<h1 class="h2-style">` and variants are a hard no from me.
- Even postfixing `-style` seems wrong.
- Why do we have 2 variants of subtitle/body/captions?
- There are so many changes, but no screenshots of what changed. I know it would be a big job to explain all with screenshots, but it's impossible to know what we are approving here.

Question: `carousel_item.jet` is deleted - not needed or mistake?

The explanation of the bigger picture should have happened prior to the PR landing. It's fine pushing opinonated strategies, in fact encouraged. If anything, that's why you were hired: for your expertise and opinions, however we need an opportunity to shape or reject opinions with reason.

I recognize the existing approach has it's issues and very open to working out the steps to a better place. This might be too much work for one PR, perhaps should be broken down a page at a time. I like the idea of components, but the atomic design chemistry metaphor breaks down for me.

jacobtoye:
I love the fact that you guys are thinking how can we make things better. I think if we make these types of changes a "bad" thing or put them in the too hard basket then you just end up treading water and never push the solution forward. That said I would suggest this type of change is quite large and sounds like it would warrant a good elab to discuss and socialize the change.

One thing I had hoped for after the elab around design systems was some momentum towards more targeted changes to support better design. The templates and relish along with our really advanced customer needs makes our situation really unique, but I reckon we have enough smart people to forge a way forward.

Maybe I'm making this bigger than it should be, but I reckon you could start the conversation rolling with an elab @matthewthomsonnz. If I am way off then "nothing to see here, carry on".

PeculiarGoat:
@matthewthomsonnz I want to revise my previous comment about OOCSS after doing some more reading...  those typography classes actually look like Atomic (functional) CSS.  And since you mentioned Atomic Design, I want to point out that Atomic CSS is unrelated to Atomic Design and IMO incompatible.  [This](https://www.browserlondon.com/blog/2019/06/10/functional-css-perils/) is a good article about some problems with Atomic CSS but it has some good points that apply to our specific case: we are using a templating system, we are using a component-based approach (gradually moving towards something like Atomic Design), we already have a huge amount of component-style CSS.

stajs:
That Zeldman quote tho

jacobtoye:
> [This](https://www.browserlondon.com/blog/2019/06/10/functional-css-perils/) is a good article about some problems with Atomic CSS but it has some good points that apply to our specific case

Not disagreeing, I think you are absolutely right when looking at our specific usecase. I just want to add that functional CSS is really amazing when used in situations where it was designed for. I.e. modern frontend design where you CSS is usually tied directly to individual components and not stored in separate CSS files. At the risk of just throwing in another opinionated article [here](https://adamwathan.me/css-utility-classes-and-separation-of-concerns/) is one by the guy that created TailwindCSS.

matthewthomsonnz:
@jacobtoye I'm crashing deliveries standup today to elab this - I'll ask about the wider path forward.

> @matthewthomsonnz I want to revise my previous comment about OOCSS after doing some more reading... those typography classes actually look like Atomic (functional) CSS. And since you mentioned Atomic Design, I want to point out that Atomic CSS is unrelated to Atomic Design and IMO incompatible.

@PeculiarGoat Riddle me this. Is my strategy different from say including the bootstrap 4 library? I know you're moving away from bootstrap, but I believe this demonstrates how it's compatible, and even fundamental, as a solid foundation.

PeculiarGoat:
@matthewthomsonnz we're moving away from bootstrap it makes things harder...

site/styles/_typography.scss
PeculiarGoat: Note to self: we decided that this isn't a huge deal in the elab.

PeculiarGoat:
@jacobtoye I just got around to reading that Adam Wathan article and wish I could unread it lol... it seems like he is so hellbent on writing less CSS that he forgot about making things understandable by humans.

Check out this example I cherrypicked from the Tailwind website:
```
<main class="py-6 px-4 sm:p-6 md:py-10 md:px-8">
  <div class="max-w-4xl mx-auto grid grid-cols-1 lg:max-w-5xl lg:gap-x-20 lg:grid-cols-2">
    <div class="relative p-3 col-start-1 row-start-1 flex flex-col-reverse rounded-lg bg-gradient-to-t from-black/75 via-black/0 sm:bg-none sm:row-start-2 sm:p-0 lg:row-start-1">
      <h1 class="mt-1 text-lg font-semibold text-white sm:text-slate-900 md:text-2xl dark:sm:text-white">Beach House in Collingwood</h1>
      <p class="text-sm leading-4 font-medium text-white sm:text-slate-500 dark:sm:text-slate-400">Entire house</p>
    </div>
    <div class="grid gap-4 col-start-1 col-end-3 row-start-1 sm:mb-6 sm:grid-cols-4 lg:gap-6 lg:col-start-2 lg:row-end-6 lg:row-span-6 lg:mb-0">
      <img src="/beach-house.jpg" alt="" class="w-full h-60 object-cover rounded-lg sm:h-52 sm:col-span-2 lg:col-span-full" loading="lazy">
      <img src="/beach-house-interior-1.jpg" alt="" class="hidden w-full h-52 object-cover rounded-lg sm:block sm:col-span-2 md:col-span-1 lg:row-start-2 lg:col-span-2 lg:h-32" loading="lazy">
      <img src="/beach-house-interior-2.jpg" alt="" class="hidden w-full h-52 object-cover rounded-lg md:block lg:row-start-2 lg:col-span-2 lg:h-32" loading="lazy">
    </div>
    <dl class="mt-4 text-xs font-medium flex items-center row-start-2 sm:mt-1 sm:row-start-3 md:mt-2.5 lg:row-start-2">
      <dt class="sr-only">Reviews</dt>
      <dd class="text-indigo-600 flex items-center dark:text-indigo-400">
        <svg width="24" height="24" fill="none" aria-hidden="true" class="mr-1 stroke-current dark:stroke-indigo-500">
          <path d="m12 5 2 5h5l-4 4 2.103 5L12 16l-5.103 3L9 14l-4-4h5l2-5Z"  stroke-width="2" stroke-linecap="round" stroke-linejoin="round" />
        </svg>
        <span>4.89 <span class="text-slate-400 font-normal">(128)</span></span>
      </dd>
      <dt class="sr-only">Location</dt>
      <dd class="flex items-center">
        <svg width="2" height="2" aria-hidden="true" fill="currentColor" class="mx-3 text-slate-300">
          <circle cx="1" cy="1" r="1" />
        </svg>
        <svg width="24" height="24" fill="none" stroke="currentColor" stroke-width="2" stroke-linecap="round" stroke-linejoin="round" class="mr-1 text-slate-400 dark:text-slate-500" aria-hidden="true">
          <path d="M18 11.034C18 14.897 12 19 12 19s-6-4.103-6-7.966C6 7.655 8.819 5 12 5s6 2.655 6 6.034Z" />
          <path d="M14 11a2 2 0 1 1-4 0 2 2 0 0 1 4 0Z" />
        </svg>
        Collingwood, Ontario
      </dd>
    </dl>
    <div class="mt-4 col-start-1 row-start-3 self-center sm:mt-0 sm:col-start-2 sm:row-start-2 sm:row-span-2 lg:mt-6 lg:col-start-1 lg:row-start-3 lg:row-end-4">
      <button type="button" class="bg-indigo-600 text-white text-sm leading-6 font-medium py-2 px-3 rounded-lg">Check availability</button>
    </div>
    <p class="mt-4 text-sm leading-6 col-start-1 sm:col-span-2 lg:mt-6 lg:row-start-4 lg:col-span-1 dark:text-slate-400">
      This sunny and spacious room is for those traveling light and looking for a comfy and cosy place to lay their head for a night or two. This beach house sits in a vibrant neighborhood littered with cafes, pubs, restaurants and supermarkets and is close to all the major attractions such as Edinburgh Castle and Arthur's Seat.
    </p>
  </div>
</main>
``` 

- At a glance, how the hell am I supposed to know what this component _is_?
- Am I really expected to memorise what all 1 million utility classes do?  If I haven't memorised them, maintaining a mental model while looking up 10+ classes to understand how one element is styled sounds like a nightmare!  How can that possibly be preferable to looking up a single component-style class?

Kazetsukai:
Just to jump on the bandwagon, that article has convinced me about the compositional aspect. I think the cherry picked example is tough to parse because what a grid layout is doing can be tough to parse. I like the general idea of abstracting out chunks of CSS into 'nuggets' that can be composed together, and really like the idea of writing CSS that doesn't need to be aware of HTML structure. I'll probably give that style a try at some point, cheers @jacobtoye. Also I haven't looked at this PR at all just got interested by the article, so I'll peace out now.

PeculiarGoat:
the-simpsons-principal-skinner

stajs:
It can elicit a visceral reaction - as HTML artisans, stylistically it is pretty ugly. But there are benefits to shifting complexity away from stylesheets to HTML.

PeculiarGoat:
Aren't stylesheets the right place for style complexity though?  What problem is solved by moving that into HTML?

stajs:
At the risk of appearing dismissive, I think the linked article makes a bunch of good points (as well as many other articles on the subject). We ought to reign in the discussion for the PR and consider a beers event to discuss. I don't think this PR is the place we choose to make the switch.

jacobtoye:
@PeculiarGoat I had the exact same reaction to tailwindcss when I first came across it. I was like wtf that even makes my javascript/jsx look bad as the lines are so long. However when actually using it I changed my mind.

From the html you posted, you would never actually have a component like that. it's doing a million different things which present a larger problem than CSS. React has soem good docs on ["thinking in react"](https://beta.reactjs.org/learn/thinking-in-react), which is applicable to all dev. It's basically breaking things down to smaller components. I.e. the S from SOLID.

So the in html you posted you would extract pieces out to reusable components. E.g. 

```js
const Title = ({ children }) => (
  <h1 class="mt-1 text-lg font-semibold text-white sm:text-slate-900 md:text-2xl dark:sm:text-white">{children}</h1>
)
```

I picked the easiest one but the technique remains the same for others. So when you do this you component becomes way easier to understand as you just look at the jsx.

```js
<Page>
  <Hero title="title" subTitle="sub title" ....
  ...
</Page>
```

I think the difference here is I'm thinking about CSS as just a part of how components are defined and the styles are tightly coupled to the components. When the css is more separated. Like maybe reusing classes elsewhere it becomes less ideal to use things like tailwindcss. Hence, why it's probably not a great idea in the Shift72 app's case. Where styles can be separated from the components (in another repo no less!).

> Am I really expected to memorise what all 1 million utility classes do?

Check out the [docs for TailwindCSS](https://tailwindcss.com/docs/configuration), press Cmd+K and search for anything. The search is sooooo good. Better than using Google + MDN.

I actually found it easier to use than raw CSS. Since the classes are basically named after the CSS rules I found it way easier. E.g. you want to add some padding to the left and right? Just add `px-4` or padding on the x axis of size 4. Rather than `padding-left: 1rem; padding-right: 1 rem;`.

The sizing takes a bit to get used to but in my experience it's a million times better than using pixels. The [spacing scale](https://tailwindcss.com/docs/customizing-spacing#default-spacing-scale) utilizes rem. So and `-4` will be 1rem. So `w-4` (width 1rem) or `pb-8` (padding bottom 2rem). This makes it really easy to sync with design as they can use a common scale. Also like all theme type things it makes change really easy. 

With a bit of experience I found it just as easy to visualise what the component looks like just as easy as when looking at CSS.

E.g.

```html
<div class="max-w-4xl mx-auto grid grid-cols-1 lg:max-w-5xl lg:gap-x-20 lg:grid-cols-2">
```

Here is a centered (`mx-auto`) 1 column `grid` with a max size of `4xl`. For all larger screens (laptop+) its bigger, has 2 columns and a gap between items. Traditionally we would have the html as `<div class="card">`. For this I would need to look in the CSS file and find the class then read the CSS rules.

---

Oh boy... sorry about the essay. I've just got a fever and the only cure is more TailwindCSS ;)

As much as I love it, I still don't think functional CSS is the right fit for our app. At least not the styles outside of relish. If you need to customize the styles in other places it's always going to be tricky.

jacobtoye:
> I don't think this PR is the place we choose to make the switch.

Sorry I missed this comment! Good shout. Sorry @matthewthomsonnz! But shout out to your for creating a good platform for this discussion even if it has polluted your PR 😬

matthewthomsonnz:
Have adjusted approach as per elab outcome. @stajs has advised to mark this ready for review.

Thanks @PeculiarGoat @smitsanghvi21 for the feedback. Ready for approval or more changes please.

matthewthomsonnz:
Also cheers @Kazetsukai @jacobtoye for the comments. IMO polluting PRs with discussion should be encouraged as its good long term.

smitsanghvi21 (approved):
Nice work, looks good to me. Would be good to wait for others to review before merging though.

stajs (approved):
Some of the font sizes are a pixel out from the designs - I think you said you did some eyeballing and Zeplin doesn't give accurate sizes? If you can give a verbal sign off that it's been past design then :shipit:

site/styles/_pages.scss
```diff
@@ -297,18 +276,11 @@
...
-    p {
-      font-size: 12px;
-      @include media-breakpoint-up(md) {
-        font-size: 16px;
-      }
     }
   }
 }
+
+.acceptinvite-invalid-token {
```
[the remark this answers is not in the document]
matthewthomsonnz: From `<s72-acceptinvite-form/>` on the acceptinvite.html.jet page.

I see `.s72-userdevices-device` for example is styled in this file but nested. I could do the same.. but i would seem unnecessary?

matthewthomsonnz:
> Some of the font sizes are a pixel out from the designs - I think you said you did some eyeballing and Zeplin doesn't give accurate sizes? If you can give a verbal sign off that it's been past design then :shipit:

@stajs A while back I asked how we loop design in at the PR stage. The answer was they're happy to view it after merge when its deployed to staging.

With that said, all good to merge?

site/styles/_pages.scss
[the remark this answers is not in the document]
stajs: It's a weird one. There's something to be said for consistency, but it's fine. It does strike me that we don't have an easy mapping of styles back to Relish components but that's a thought for another day.

stajs:
> With that said, all good to merge?

Since you asked so politely, sure. We're talking 'bout some pixels so not the end of the world, but you could have deployed this branch to staging for that design walk through and got it signed off prior to merging.

matthewthomsonnz:
> > With that said, all good to merge?
> 
> Since you asked so politely, sure. We're talking 'bout some pixels so not the end of the world, but you could have deployed this branch to staging for that design walk through and got it signed off prior to merging.

@stajs Have uploaded to staging-event and messaged bex for approval

site/styles/_typography.scss
```diff
@@ -0,0 +1,80 @@
+@mixin typography-style($font-size, $letter-spacing:null, $font-weight:null, $margin-bottom:null, $font-color:null, $line-height:null) {
+  font-size: $font-size;
+  letter-spacing: $letter-spacing;
+  font-weight: $font-weight;
+  margin-bottom: $margin-bottom;
+  color: $font-color;
```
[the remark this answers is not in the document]
matthewthomsonnz: Nope sass just removes them. Good catch though

jacobtoye:
> Have uploaded to staging-event and messaged bex for approval

Just want to give you a shout out, as this is the perfect thing to do.

matthewthomsonnz:
@stajs @PeculiarGoat The following design changes have been made since PR was approved:

- use rem instead of px for accessibility (added a function that converts pixels)
- adjust mobile heading 1

Let me know if the pixel conversion function is all good with you and I'll merge

site/styles/_typography.scss
```diff
@@ -0,0 +1,84 @@
+@function to-rem($pixels) {
+  @return ($pixels / 20) * 1rem;
+}
+
+@mixin typography-style($font-size, $letter-spacing:null, $font-weight:null, $margin-bottom:null, $font-color:null, $line-height:null) {
+  font-size: to_rem($font-size);
```
PeculiarGoat, line 6: Should that be `to-rem` not `to_rem` ?

site/styles/_typography.scss
```diff
@@ -0,0 +1,84 @@
+@function to-rem($pixels) {
```
PeculiarGoat, line 1: I'm confused about what this is for... wouldn't it be better to fully commit to using `rem` or `px` instead of inputting one and outputting the other?

PeculiarGoat (changes requested):
Couple of things.  Also, can you please merge develop into this branch and run `npm run css:lint`.

site/styles/_typography.scss
matthewthomsonnz: doh. changed

site/styles/_typography.scss
matthewthomsonnz: `rem` feels abstract and unintuitive e.g. `2.8125rem; 1.4375rem; 0.875rem` even if you know what the base size is. I think its good to describe the actual value it translates/computes to. Also makes it easier if design wants to change the value.
```
@mixin heading-1-style() {
  @include typography-style(2.8125rem, 0.4px, $font-weight-normal, 8px, var(--body-color)); // 46px
```
Will change if you prefer though I don't mind

matthewthomsonnz:
merged and linted

site/styles/_typography.scss
stajs: https://stackoverflow.com/questions/30374863/why-use-rem-instead-px-when-its-the-same-anyway

Using `rem` has a major accessibility advantage for people with poor eyesight that have increased their browser `rem` base size from the default of `16px`. A function that hardcodes division by `16` is going to screw them over. Yes, I agree that the floating point `rem` values are abstract and unintuitive and I like that you tried to improve on that, but this implementation is not a good solution.

But having said all that, I'm of the opinion that anyone who has increased their browser base size is going to have a poor result with our site anyway with a mix of `rem` and `px` and a design that is trying to be pixel perfect.

stajs (changes requested):

site/styles/_typography.scss
PeculiarGoat: I think we should stick with all `px` for now.  There is probably value in converting things to `rem` and `em`, but it's something that we should have an elab for and it will affect more than just fonts so I reckon this PR is not the right place for it.

PeculiarGoat (changes requested):
I didn't pick up on these problems before... those `h` elements had a specific hierarchy for screen readers and they will need to be changed back.

You can use the [WAVE extension](https://wave.webaim.org/extension/) (turn it on and click the structure tab) to view the hierarchy.  It's important to avoid skipping numbers in the hierarchy, and it's especially important that each page only has one `h1` element... not just for screen readers but it's also helpful for SEO!

site/templates/collection/carousel/item/title.jet
```diff
@@ -1,3 +1,3 @@
 {{block carouselItemTitle(title)}}
-  <h3>{{title}}</h3>
+  <h1>{{title}}</h1>
```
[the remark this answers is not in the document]
matthewthomsonnz: - Elem changed to h3 and carousel h3 css replaced with heading-1 mixin

site/templates/collection/wishlist.jet
```diff
@@ -32,7 +32,7 @@
       {{end}}
             {{if i18n("userwishlist_slider_header") != ""}}
             <div class="swiper-title">
-                <h3>{{i18n("userwishlist_slider_header")}}</h3>
+                <h2>{{i18n("userwishlist_slider_header")}}</h2>
```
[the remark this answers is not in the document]
matthewthomsonnz: - changed

site/templates/collection/slider.jet
```diff
@@ -7,7 +7,7 @@
     <div class="swiper-container" data-items-per-row="{{.ItemsPerRow}}" data-layout="{{.ItemLayout}}">
       {{if .DisplayName != ""}}
       <div class="swiper-title">
-          <h3>{{.DisplayName}}</h3>
+          <h2>{{.DisplayName}}</h2>
```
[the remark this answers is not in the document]
matthewthomsonnz: - changed

site/templates/items/meta_item.jet
```diff
@@ -90,7 +90,7 @@
 
   {{if displayTitleAndGenreBelowPosters}}
     <div class="caption">
-      <div class="title" title="{{item.Title}}">{{title|stripHTML}}</div>
+      <h3 class="title" title="{{item.Title}}">{{title|stripHTML}}</h3>
```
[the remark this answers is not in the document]
matthewthomsonnz: - changed and css updated to target h4 instead of h3

site/templates/page/partial.jet
```diff
@@ -23,7 +23,7 @@
 
   {{ if config("media_item_caption") == "true" }}
     <div class="caption">
-      <div class="title" title="{{page.Title}}">{{page.Title}}</div>
+      <h4 class="title" title="{{page.Title}}">{{page.Title}}</h4>
```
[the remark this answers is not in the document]
matthewthomsonnz: - changed to h2, css adjusted

matthewthomsonnz:
> I didn't pick up on these problems before... those `h` elements had a specific hierarchy for screen readers and they will need to be changed back.
> 
> You can use the [WAVE extension](https://wave.webaim.org/extension/) (turn it on and click the structure tab) to view the hierarchy. It's important to avoid skipping numbers in the hierarchy, and it's especially important that each page only has one `h1` element... not just for screen readers but it's also helpful for SEO!

@PeculiarGoat No problem - changes made. Can I merge this now please?

That extensions pretty cool I'll use it going forward.

site/styles/_typography.scss
stajs: @matthewthomsonnz What is the resolution on this one? We are agreeing to stick with `px` for now right, so `to-rem()` should be `rem`oved?

stajs (changes requested):

site/styles/_typography.scss
matthewthomsonnz: @stajs Oh yep. Reverted file to before function was added. 

iseewhatyoudidthere.jpg

stajs (approved):
I think it's good, pending @PeculiarGoat.

site/templates/collection/partial.jet
```diff
@@ -13,6 +13,6 @@
   </a>
 
   <div class="caption">
-    <div class="title" title="{{collection.Title}}">{{collection.Title}}</div>
+    <h4 class="title" title="{{collection.Title}}">{{collection.Title}}</h4>
```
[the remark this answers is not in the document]
PeculiarGoat: Yeah I reckon it's a relic too.  Happy to just ignore it for this PR.

PeculiarGoat (changes requested):
There is an issue with those partial caption titles.  The form pages (wishlist, library, search) expect `h2`, but the same partials are used in the homepage wishlist slider which expects `h4`.  I think the easiest solution to this is just to go back to `div` for those titles... otherwise I'm not sure how we can handle it.

Everything else is looking okay though.

matthewthomsonnz:
> There is an issue with those partial caption titles. The form pages (wishlist, library, search) expect `h2`, but the same partials are used in the homepage wishlist slider which expects `h4`. I think the easiest solution to this is just to go back to `div` for those titles... otherwise I'm not sure how we can handle it.
> 
> Everything else is looking okay though.

- Have changed those h4 partials to div
- Updated style to target those .title divs

PeculiarGoat:
@matthewthomsonnz can you make all of the partial titles divs instead of a mixture of h2 and div for consistency.  I think we should look at how to fix this issue in a separate task.

matthewthomsonnz:
> @matthewthomsonnz can you make all of the partial titles divs instead of a mixture of h2 and div for consistency. I think we should look at how to fix this issue in a separate task.

@PeculiarGoat done.

site/templates/collection/partial.jet
```diff
@@ -13,6 +13,6 @@
   </a>
 
   <div class="caption">
-    <div class="title" title="{{collection.Title}}">{{collection.Title}}</div>
+    <div class="title" title="{{collection.Title}}">{{collection.Title}}</hdiv>
```
PeculiarGoat, line 16:
```suggestion
    <div class="title" title="{{collection.Title}}">{{collection.Title}}</div>
```

PeculiarGoat (approved):
Let's goooooooo!

PR #370, opened 2022-06-14
Tweak:  change CanBeWatchedButton font weight to match play button

sam-shift72:
Not sure if it's intentional, but the can-be-watched button has a heavier font weight than the play button next to it.

This PR makes it normal weight to match the regular button styling. It seemed like this might have just been some copypasta styles that got overlooked when the button styling was changed.

Please close if this was the design intent.

Before:
image

After:
image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 ### Changed
 - Awards icon from star to laurel
 - Show play and pricing buttons based on item type on `meta_item.jet`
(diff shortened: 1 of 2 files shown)


PeculiarGoat (approved):
LGTM.  Could've sworn I fixed that recently.

l0ud0gg (approved):
👍 nice, thanks for that

PeculiarGoat:
@sam-shift72 Could you put a note in the changelog too please!

PR #372, opened 2022-06-15
AB#7907 Credit Card icon alignment

twln1:
ADO card: ☑️ [AB#7907](https://dev.azure.com/S72/SHIFT72/_workitems/edit/7907)

## Description of work
Fix the CC icon so it isn't cut off image

- saved_cards

## Checklist


PeculiarGoat (changes requested):
I've tested this out and I feel that this is a bandaid.  Changing the padding like you have also affects the list of saved cards... which TBH doesn't look bad but it's a side effect and it's because you are sidestepping the real problem here.

The real problem is:
- All of our credit card brand glyphs overflow their EM square in the woff file.
- We also have `overflow: hidden` on the `label` that contains this icon.

I would like to keep `overflow: hidden` on the `label` so we can still have that ellipsis on the right.  But I would like the glyphs to be fixed in the woff rather than hacking in solutions with CSS.

Also worth noting, the other glyphs in that file appear to have inconsistent baselines... that should probably be tidied up.  And if you compare with a Font Awesome font for example, you will see that they don't even use baselines... it looks like they try to fill the EM square as much as possible.  Is there a reason we haven't done it that way?

smitsanghvi21:
Yeah I also agree with Jack's comment that it also affects the list of saved cards. But I feel this isn't very noticeable and the temporary solution is good enough to deal with this. Maybe this can be worked upon later on in the future as part of a bigger work related to modal and fonts.

smitsanghvi21 (commented):
BTW Haven't checked how it looks  on smaller devices

twln1:
Fair enough. I thought the padding was fairly inconsequential.
100% agree on the overflow.

Instead how about some different SVGs that work better for what we're trying to achieve. Thanks @matthewthomsonnz 
Had to bump the font size a little otherwise it's noticeably smaller than the old icons

Before image

After  image

PeculiarGoat (approved):
@twln1 yup I can see the glyph overflow fixed in the woff... much better!  The baselines are still inconsistent (across all glyphs, not the credit cards specifically) but that can be a separate task... I guess we would need to do an elab with design team and figure out some iconography rules.

PR #378, opened 2022-06-17
Line clamp on carousel heading

curtisadams28:
## Description of work
- Adds a mixin that adds the line-clamp property with the number of lines as an argument.
- Line clamped the carousel heading to 3 lines.

## Testing
- Create a heading that is long enough that it should wrap to 4 lines.

## Checklist
- [x] Design review


PeculiarGoat:
Is it worth adding a mixin that only gets used once?

smitsanghvi21 (commented):
Upon having a brief look couldn't really see a visual difference on how the heading is rendering after/before. Not sure If I am missing something.

PeculiarGoat:
Is there an actual plan to use it?  If not, then we can probably live without the additional complexity for now.

stajs:
Will the film detail page need it?

curtisadams28:
@stajs Maybe. A long title could push the rest of the content below the fold. I've removed the mixin for now and it can be added back later if we need it.

PeculiarGoat:
Have you checked those properties on caniuse?

curtisadams28:
Yes. I've also tested in Chrome and Firefox. Still need to test on IOS.

stajs:
[Support is good for `-webkit-line-clamp`](https://caniuse.com/mdn-css_properties_-webkit-line-clamp), suprisingly so even for non-Webkit browsers.

If you look at [the non-prefix "official" standard `line-clamp`](https://caniuse.com/css-line-clamp), they push you to the prefix:

image

brynyShift72:
Apart from iOS is that everything that needs to be checked?
Are you wanting to merge this with 1.0.1 or after?

site/styles/_carousel.scss
```diff
@@ -131,6 +131,13 @@ s72-carousel {
     max-width: 930px;
     padding-bottom: 0;
 
+    /* stylelint-disable */
+    display: -webkit-box;
+    overflow: hidden;
+    -webkit-line-clamp: 3;
+    -webkit-box-orient: vertical;
```
PeculiarGoat, line 138: Do we care that `box-orient` is deprecated?

PeculiarGoat (changes requested):
I merged in develop + ran the linter.

site/styles/_carousel.scss
stajs: I care, but MDN is the one reporting it deprecated, and yet [use it in their own example](https://developer.mozilla.org/en-US/docs/Web/CSS/-webkit-line-clamp). 🤷

stajs:
@curtisadams28 just checking: @PeculiarGoat requested changes, not sure what they were but you had confirmed it was okay?

stajs:
@curtisadams28 just checking: @PeculiarGoat requested changes, not sure what they were but you had confirmed it was okay?

PR #386, opened 2022-06-29
Detail page mobile - CTA button adjustments

matthewthomsonnz:
ADO card: ☑️ [AB#8093](https://dev.azure.com/S72/SHIFT72/_workitems/edit/8093)

Designs: 
- 📱 [Mobile](https://app.zeplin.io/project/5fdaa165fbacf584a2ce4f8b/screen/62030a7ff375ab015a507fad)

## Description of work
- ~Makes detail page pricing buttons full width on xs breakpoint~ this PR is now just translations
- Changes translation "Play now" to "Watch now"

### Affected Clients
 - All new clients who use Kibble

## Checklist
- [ ] Design review

### DIFF
--- a/site/styles/_meta-detail.scss
+++ b/site/styles/_meta-detail.scss
@@ -283,6 +283,19 @@ s72-element-switcher,
       font-size: 12px;
     }
   }
+
+  .s72-pricing-button-container {
+    display: flex;
+    flex-wrap: wrap;
+
+    @include media-breakpoint-up(xs) {
+      display: grid;
+    }
+
+    .s72-subscribe {
+      flex-grow: 1;
+    }
+  }
 }
 
 .recommendations-collection {
--- a/site/styles/_meta-detail.scss
+++ b/site/styles/_meta-detail.scss
@@ -240,19 +240,6 @@ s72-element-switcher,
       animation: fadein 2s;
     }
   }
-
-  .s72-pricing-button-container {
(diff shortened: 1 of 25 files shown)


site/styles/_meta-detail.scss
```diff
@@ -240,6 +240,19 @@ s72-element-switcher,
       animation: fadein 2s;
     }
   }
+
+  .s72-pricing-button-container {
```
PeculiarGoat, line 244: Bex is wrong about this... you should delete it IMO.

- It isn't more visually appealing... it looks cramped.
- It isn't consistent... these buttons should look the same wherever they are.
- You can't use `flex` here unless you also override to unset `gap` (because Safari sucks)... but doing it that way is CSSpaghetti... you should just use `grid`.
- This override should live in `_cta-buttons.scss`... one thing should live in one place.  (though in this case I prefer if it didn't exist at all).

PeculiarGoat (changes requested):
Translations are probably fine... hope we don't get clients asking to change it back.

I'm not keen on that CSS though!

site/styles/_meta-detail.scss
matthewthomsonnz: Asked Bex and have reverted css, this pr is now just translations

PeculiarGoat (approved):
👍

PR #400, opened 2022-07-20
Add font size vars

smitsanghvi21:
Due to the recent typography changes some of the variables were removed which allowed to adjust the font size of the content. This PR is adding vars to change font size of headings, body text, etc.


curtisadams28 (changes requested):
You added the mixins but didn't remove the old ones with the same names. Won't that cause issues?

smitsanghvi21:
@curtisadams28 Hmm not sure what do you mean? I have just created vars and passed on to the existing mixins? Also haven't added new mixins.

curtisadams28:
@smitsanghvi21 `@mixin subtitle-1-style()` exists twice on line 75 and 98.

site/styles/_variables.scss
```diff
@@ -258,3 +258,8 @@ $font-family-base: 'Poppins', system-ui, -apple-system, 'Segoe UI', Roboto, 'Hel
 $font-weight-light: 300 !default;
 $font-weight-normal: 400 !default;
 $font-weight-bold: 500 !default;
+$font-size-body-1-style: 16px !default;
```
PeculiarGoat, line 261: These need to be CSS variables in `:root`.

site/styles/_typography.scss
```diff
@@ -65,6 +65,24 @@
 }
 
 @mixin body-1-style() {
+  @include typography-style($font-size-body-1-style, 0.5px, $font-weight-light, null, null, 23px);
+}
+
+@mixin body-2-style() {
```
PeculiarGoat, line 71: As @curtisadams28 said... you don't need to redefine all of these mixins.

PeculiarGoat (changes requested):
Some changes to look at!

curtisadams28:
@PeculiarGoat @smitsanghvi21 Do you think it would be worth adding variables for the letter spacing too? By setting a value, they've deviated from the default letter spacing intended by the maker of the font. Different fonts have different letter spacings so I'm concerned that the spacing won't look right for custom fonts. It would be nice to set something like this `--font-letter-spacing: normal;`

site/styles/_variables.scss
```diff
@@ -50,6 +50,12 @@
   --page-detail-padding-top: 120px;
   --page-detail-padding-top-md: 120px;
   --page-detail-padding-top-lg: 180px;
+
+  --font-size-body-1-style: 16px;
```
PeculiarGoat, line 54: If we are going to have several exposed variables per mixin, it might be easier to read if you have the property name after the mixin name e.g. `--body-1-style-font-size: 16px;`

PeculiarGoat: Also, it's probably worth exposing the sizes for all of the mixins (you didn't do the headings or overline mixins).

smitsanghvi21: I think `--font-size-body-1-style` looks more readable than `--body-1-style-font-size`

PeculiarGoat: This is much better:
```
--body-1-style-letter-spacing: normal;
--body-1-style-font-size: 16px;
--body-2-style-letter-spacing: normal;
--body-2-style-font-size: 24px;
```

Than this:
```
--letter-spacing-body-1-style: normal;
--font-size-body-1-style: 16px;
--letter-spacing-body-2-style: normal;
--font-size-body-2-style: 24px;
```

smitsanghvi21: @PeculiarGoat The headings one has got different font sizes for various breakpoints. I think it's going to just complicate if we end up ending 10 vars for just the heading

PeculiarGoat: It's going to get complicated if we expose some font sizes but not others... they should all be exposed.

smitsanghvi21: But some of the headings are never going to get changed like the `small` in `heading-1`. Are you suggesting of adding a var for `heading-1-style`, `heading-1-style-desktop`, `heading-1-style-small`, etc?

PeculiarGoat: Yes please.  But they should have the breakpoint labels rather than `desktop`, `tablet`, etc.

PeculiarGoat (approved):
Looks good.  Don't forget the changelog.

Have you tested that they all work okay?

smitsanghvi21:
@PeculiarGoat Just briefly tested it and looks fine. Might want someone else to test it as well.

PR #408, opened 2022-07-29
Missing translations for SVOD

ashesnz:
ADO card: ☑️ [AB#9319](https://dev.azure.com/S72/SHIFT72/_workitems/edit/9319)

## Description of work
Theres a few translations missing for SVOD, [which we default to English](https://github.com/indiereign/shift72-web/blame/fe7a6b5f1dadc8f9030ed03d9a756195bbfdc0cc/src/transactional/subscriptions/user-subscription.component.js#L135), which isn't great for our international customers. Since we aren't using the site_api_translation yet for filmicca, I've added them here.

### DIFF
--- a/site/ar_LB.all.json
+++ b/site/ar_LB.all.json
@@ -988,9 +988,6 @@
   "usersubscriptions_subscription_status_cancelled": {
     "other": "ألغيت"
   },
-  "usersubscriptions_subscription_status_errored": {
-    "other": "usersubscriptions_subscription_status_errored"
-  },
   "usersubscriptions_subscription_status_past_due": {
     "other": "فات موعد الاستحقاق"
   },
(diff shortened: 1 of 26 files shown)


brynyShift72 (commented):
LGTM, wondering if there shouldn't be translations where the value is the key though

brynyShift72 (approved):

PeculiarGoat (approved):
👍

ashesnz:
> LGTM, wondering if there shouldn't be translations where the value is the key though

Agreed, removed `usersubscriptions_subscription_status_errored` as couldn't see it being used anywhere in relish

stajs:
@ashesnz FYI filmicca _are_ using DB translations, so you'll have to update the DB if you haven't already.

ashesnz:
> @ashesnz FYI filmicca _are_ using DB translations, so you'll have to update the DB if you haven't already.

Thanks for the headsup! Soz was looking at staging while testing, looks like they are those keys are in prod 👍🏽

PR #427, opened 2022-08-25
Ab#9361 Translations and css availability label live events

matthewthomsonnz:
ADO card: ☑️ [AB#9361](https://dev.azure.com/S72/fefacba9-96b6-4af2-a53d-050ed453e13a/_workitems/edit/9361)

Elab notes/AC: 📃 [Google Docs](https://docs.google.com/document/d/1xVECgmuo5F5HxacMQQUDC3wXsbcCdOQyExJbL28B00Y/edit?pli=1#heading=h.4u4auh6hkf7w)
- [ ] Has this been discussed with Delivery Team?

Designs: 
- 🖥 [mobile/tablet/desktop](https://app.zeplin.io/project/607f961cf9f03790cc66e475/screen/63044e6e8388db11c0f51a35)

## Description of work
All translations needed for live player except the shopping modal blurb

## Checklist
- [x] CI tests are passing Github actions (inc. linting)
- [ ] Key areas of the feature outlined for context and testing
- [x] If there are designs for this work are they noted here and in the ADO card 
- [ ] Design review
- [ ] Have checked this at multiple screen resolutions and range of browsers
- [ ] Moved ADO card to Dev/done, checked link to Github, tagged "Review"
- [x] Updated changelog (if applicable)
- [x] I promise to document any new feature toggles/configurations in the appropriate documentation

PR #429, opened 2022-08-31
Ab#9564 live lozenge

PeculiarGoat:
ADO card: ☑️ [AB#9564](https://dev.azure.com/S72/SHIFT72/_workitems/edit/9564)

Designs: 
- 📱 [Here](https://zpl.io/jZo78Zm)

## Description of work
- Added new live lozenge component to carousel slides and film detail pages.
- Refactored carousel slide and film detail jets + CSS with a hammer.
- Added translations for live lozenge component.

## Related PRs 
- https://github.com/indiereign/shift72-web/pull/345

## Checklist


site/styles/_availability-tags.scss
```diff
@@ -71,3 +71,25 @@ s72-plan-label {
   font-size: 12px;
   text-align: center;
 }
+
+s72-live-lozenge {
+  .live-lozenge {
+    background-color: var(--live-lozenge-background-color);
```
matthewthomsonnz, line 77: We can utilise `backdrop-filter: blur(...)` and rgba to get the blur effect in the designs

site/styles/_availability-tags.scss
```diff
@@ -71,3 +71,25 @@ s72-plan-label {
...
+    border-radius: 100px;
+    font-size: 12px;
+    line-height: 30px;
```
matthewthomsonnz, line 80: The circle isn't quite aligning vertically with the text, its appears a pixel or two below on my screen 
image

site/styles/_availability-tags.scss
```diff
@@ -71,3 +71,25 @@ s72-plan-label {
...
+
+    &.is-live {
+      background-color: var(--live-lozenge-live-background-color);
+    }
+
+    .fa-circle {
+      @extend .fa;
+
+      margin-right: 8px;
+    }
```
matthewthomsonnz, line 93: Should this be a before pseudo element mixin that makes a circle instead? e.g. `content: "", border: 50%, display: inline-block etc`

site/styles/_availability-tags.scss
PeculiarGoat: We made a decision a while back to not use opacity in core-template because of the potential accessibility issues with client colours + client background images.

site/styles/_carousel.scss
```diff
@@ -96,8 +96,7 @@ s72-carousel {
   @extend .carousel-bottom-gradient;
 
   bottom: 0;
-  display: flex;
-  flex-direction: column;
+  display: grid;
```
matthewthomsonnz, line 99: Does this change follows these principles? https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Grid_Layout/Relationship_of_Grid_Layout

site/styles/_availability-tags.scss
matthewthomsonnz: Can we confirm with product?

site/styles/_availability-tags.scss
PeculiarGoat: I've gone with a negative margin hack to bump the dot a bit higher.  I think this has something to do with our icon glyphs not fitting into their EM squares as expected.  Although... looking at what is in that glyph... I would expect it to appear too high, not too low.

site/styles/_availability-tags.scss
PeculiarGoat: Could do... I just rolled with circle icon because it's there in our custom font (`fa-` is a red herring).

site/styles/_carousel.scss
PeculiarGoat: It does if we assume that the 2-dimensional layout is necessary.  I prefer the 2-dimensional layout because it means we don't have to worry about dirty hacks with margins and absolute positions etc.

site/templates/film/item.jet
```diff
@@ -27,16 +27,22 @@
...
-              <s72-image src="{{film.ImageMap["Portrait"]}}" alt="{{film.Title}}" class="poster poster-image poster-image-portrait"></s72-image>
-            </div>
+
+            {{if film.ImageMap["Portrait"] != ""}}
+              <div class="poster poster-portrait">
+                <s72-availability-status data-slug="{{film.Slug}}"></s72-availability-status>
+                <s72-image src="{{film.ImageMap["Classification"]}}" alt="" class="classification-image">
+                <s72-image src="{{film.ImageMap["Portrait"]}}" alt="{{film.Title}}" class="poster poster-image poster-image-portrait"></s72-image>
+              </div>
+            {{end}}
```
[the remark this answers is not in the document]
PeculiarGoat: It's related to the switch to grid... if a film has no poster image then these wrappers would prevent the page content from shunting to the left as grid would consider the div to be content.

site/styles/_availability-tags.scss
PeculiarGoat: Switched to blur though it doesn't look very good IMO.

site/en_AU.all.json
```diff
@@ -1412,5 +1412,20 @@
...
+    "other": "LIVE"
+  },
+  "live_lozenge_minutes": {
+    "other": "LIVE: {{.Minutes}} MINUTES"
+  },
+  "live_lozenge_hours": {
+    "other": "LIVE: {{.Hours}} HOURS"
+  },
+  "live_lozenge_days": {
+    "other": "LIVE: {{.Days}} DAYS"
```
additiverich, line 1429: Just want to check we're okay with these ALL IN CAPS by default and not just converting them to caps in CSS. We are using `text-transform: uppercase` elsewhere in our stylesheets and I feel we should probably be consistent.

additiverich (commented):
The references to 'lozenge' might need to change as well

site/styles/_carousel.scss
matthewthomsonnz: I think its incorrect that we're considering negative margins and absolute positions as dirty hacks. A quick google finds consensus that they're not bad practice and I consider them the most suitable approach in many scenarios.

We should use grid for grid layouts, and reconsider if its good practice otherwise.

site/templates/film/item.jet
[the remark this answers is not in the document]
matthewthomsonnz: I see what you mean. These changes should be separate though if its not related to make reviewing easier.

site/styles/_availability-tags.scss
matthewthomsonnz: Not trying to be contentious but I feel like this is a better solution:
```
display: flex;
align-items: center;

**Comment by matthewthomsonnz** [comment]
I think this PR should include display none on the detail page poster availability status

File: site/styles/_availability-tags.scss (hunk 2 above)
Comment (PeculiarGoat) [reply site/styles/_availability-tags.scss:81]: Definitely a better solution!

File: site/styles/_availability-tags.scss (hunk 3 above)
Comment (PeculiarGoat) [reply site/styles/_availability-tags.scss]: @matthewthomsonnz @additiverich Can I get some votes... custom circle or fa-circle?  Core-template has a mixture of custom made icons and fa-icons so hard to argue one way or the other based on consistency.

File: site/ar_LB.all.json [hunk 7]
```diff
@@ -1464,5 +1464,20 @@
...
     "other": "انتهت جلسة التسوق ، يرجى المحاولة مرة أخرى."
+  },
+  "live_label_ended": {
+    "other": "مباشر: انتهى"
+  },
+  "live_label_live": {
+    "other": "يعيش"
+  },
+  "live_label_minutes": {
+    "other": "البث المباشر {{.Minutes}} دقائق من الدقائق"
```
additiverich, line 1475: Don't we need more pluralities that this esp. for Arabic?

site/en_AU.all.json
additiverich: @matthewthomsonnz Do you have an opinion on this?

site/styles/_carousel.scss
PeculiarGoat: Yes it's good practice because it has two rows now.  :P

site/en_AU.all.json
PeculiarGoat: We don't do text transform in core-template... when a client wants it changed, overriding CSS in a template is more annoying than changing their translations in the db.

site/ar_LB.all.json
PeculiarGoat: Only need pluralities if there is a `{{.Count}}`.

site/en_AU.all.json
matthewthomsonnz: Uppercase translations is probably best to give clients more control in the future

IMO design was sacrificed with the strictly no transforms dictate. The issue was non-english clients that didn't want uppercase right? It would be nice to allow text transform for en only, or maybe even just the latin languages

site/styles/_availability-tags.scss
matthewthomsonnz: [I used custom circle](https://github.com/shift72/core-template/pull/427/files#diff-1690b8cb885e67b64253816006cb361c5af104c3073fe62ebfb567c91815dcd5R72-R80) and it aligned well. Either or I don't mind both solutions work.

Whichever choice can you make a reusable mixin and call it here? As I'll add it to my branch for consistency

PeculiarGoat: Mixin added!

site/ar_LB.all.json
additiverich: Doesn't `{{.Minutes}}` act as a de facto `{{.Count}}`? It's an integer which implies a different message in English at the very least (live in 1 minute vs. live in 2 minutes)

additiverich (approved):
If @matthewthomsonnz is happy with the CSS I'm happy

matthewthomsonnz (approved):

PR #430, opened 2022-08-31
Ab#9013 spacing between components

matthewthomsonnz:
ADO card: ☑️ [AB#8089](https://dev.azure.com/S72/fefacba9-96b6-4af2-a53d-050ed453e13a/_workitems/edit/8089)

## Description of work
This card implements spacing feedback from design as per annotated images in [this google doc](https://docs.google.com/document/d/1EJlDKaCQ7PIDVE3P_dPo6e765CkYVQPZro7v4m9FEgQ/edit#heading=h.51usj6fb9f8h)

### Affected Clients
 - All clients who use Kibble

## Checklist
- [x] Design review

### DIFF
--- a/site/styles/_awards.scss
+++ b/site/styles/_awards.scss
@@ -6,6 +6,7 @@
     gap: 30px 104px;
     grid-auto-rows: 1fr;
     margin-bottom: 51px;
+    margin-top: -14px;
 
     @include media-breakpoint-up(md) {
       grid-template-columns: repeat(2, minmax(0, 320px));
--- a/site/styles/_awards.scss
+++ b/site/styles/_awards.scss
@@ -13,7 +13,7 @@
     }
 
     @include media-breakpoint-up(lg) {
-      margin-bottom: 69px;
+      margin-bottom: 52px;
     }
   }
 
--- a/site/templates/tv/detail.jet
+++ b/site/templates/tv/detail.jet
@@ -31,8 +31,10 @@
         {{end}}
         <div class="meta-detail-content">
           <h1>{{tvseason.GetTitle(i18n)}}</h1>
-          <div class="meta-detail-tagline">
-            {{yield metaItemTagline() tvseason}}
+          <div class="meta-detail-tagline-and-classification">
+            <div class="meta-detail-tagline">
+              {{yield metaItemTagline() tvseason}}
+            </div>
           </div>
           <s72-availability-label data-slug="{{tvseason.Slug}}"></s72-availability-label>
 
--- a/site/styles/_availability-tags.scss
+++ b/site/styles/_availability-tags.scss
@@ -2,8 +2,8 @@ s72-availability-label {
   .availability-state {
     @extend .d-flex;
     @include body-1-style;
-    margin-bottom: 16px;
     line-height: 24px;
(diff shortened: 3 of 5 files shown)


site/styles/_awards.scss
```diff
@@ -6,13 +6,14 @@
     gap: 30px 104px;
     grid-auto-rows: 1fr;
     margin-bottom: 51px;
+    margin-top: -14px;
```
PeculiarGoat, line 9: Is there a reason why you can't remove/reduce the margin-bottom on the elements above the awards instead of using this forbidden magic?

site/templates/tv/detail.jet
```diff
@@ -31,8 +31,10 @@
         {{end}}
         <div class="meta-detail-content">
           <h1>{{tvseason.GetTitle(i18n)}}</h1>
-          <div class="meta-detail-tagline">
-            {{yield metaItemTagline() tvseason}}
+          <div class="meta-detail-tagline-and-classification">
```
PeculiarGoat, line 34: What does this do other than make the tagline elements 3 containers deep instead of 2?

PeculiarGoat (changes requested):
Looks fine apart from those bits I've commented on.

However... I think that google doc being a source of truth is questionable... shouldn't this stuff be in Zeplin?

site/templates/tv/detail.jet
matthewthomsonnz: - Adds consistency with the film page which has this div
- There's a margin that's applied via this class that appears on the film detail page (see [this bug card](https://dev.azure.com/S72/SHIFT72/_boards/board/t/Web%20team/Stories/?workitem=9218))
- Then I've adjusted the margin for both film and tv.

I often come across very similar and repeat code in the template (and css), starting to like your start from scratch idea

site/styles/_awards.scss
matthewthomsonnz: Its to satisfy this scenario 
image

matthewthomsonnz:
> Looks fine apart from those bits I've commented on.
> 
> However... I think that google doc being a source of truth is questionable... shouldn't this stuff be in Zeplin?

Definately agreed. You cant talk to the actual coded components appearance on Zeplin though. If you have any ideas it would be good to chat at standup.

site/templates/tv/detail.jet
PeculiarGoat: If this adds consistency then 👍 .

site/styles/_awards.scss
matthewthomsonnz: Nope, this problem would remain if see more/less was just a static block element at the bottom of the element switcher

PeculiarGoat: Is it supposed to look like this?
image

That gap is not the same as your screenshot.  Am I confused?

matthewthomsonnz: It is. Oh the screenshot is of the google doc, which is how it looked before. The red squares are what the gaps should be, not measurements.

PeculiarGoat: If that's how it's supposed to look then resolved!

PeculiarGoat (approved):
👍

PR #454, opened 2022-10-07
AB#7318 routeToSlug doesn't support episodes

additiverich:
ADO card: ☑️ [AB#7318](https://dev.azure.com/S72/fefacba9-96b6-4af2-a53d-050ed453e13a/_workitems/edit/7318)

## Description of work
- Episodes can be added to a collection (which is supported by the API and admin) but kibble hasn't had a pattern defined in kibble.json for the `routeToSlug` function to use to link to them.
- Historically we have used named anchors for each episode on the tv season detail page but these seem to have been disappeared recently. I've reinstated these and used them in the kibble.json `urlPath` pattern.
- Maybe ultimately a site could support dedicated episode pages but for the interim this should do the trick.

### Affected Clients
 - All clients who use Kibble and have episodes and want to link to them from collections (inc. the homepage carousel)

## Checklist


site/templates/items/meta_item.jet
```diff
@@ -31,7 +31,6 @@
   {{pillSupplement := ""}}
   {{slugRoute := routeToSlug(item.Slug)}}
   {{if isEpisode}}
-    {{slugRoute = routeToSlug(item.InnerItem.Season.Slug) + "#episode-" + item.InnerItem.EpisodeNumber}}
```
additiverich, line 34: We were making an episode-specific route here but has been delegated to kibble instead

matthewthomsonnz:
Ugh FOUC strikes again. On my end, if I go to http://localhost:8081/tv/real-tv/1#episode-1 (pointed to staging-store-kibble). It scrolls to episode one, then the components mount, pushing episode one above the screen.

It could be nice to do something like on s72 loaded smooth scrollTo.

Bonus points for adding a nice css fade highlight to a grey bg.

additiverich:
@matthewthomsonnz Just wondering when you think the page is fully loaded and could be a trigger to call scrollTo. Aren't our components redrawing all the time? If I scroll to episode 1, and then scroll away, and a modal window renders, is that a trigger to scroll to the named anchor?

matthewthomsonnz:
> @matthewthomsonnz Just wondering when you think the page is fully loaded and could be a trigger to call scrollTo. Aren't our components redrawing all the time? If I scroll to episode 1, and then scroll away, and a modal window renders, is that a trigger to scroll to the named anchor?

It should be 'DOMContentLoaded', but because we have FOUC we can't, maybe the solution is to fix the FOUC?

It appears the only other way is window onload, but isn't consistent as it tends to fire on all images loaded, which can be before the js if cached

additiverich:
@matthewthomsonnz It's those lazy loading images which are messing up the height of the page. Are we at a point where we need to do something for core-template that images take up the requisite amount of space before they're loaded.

additiverich:
Parking this until time frees up

matthewthomsonnz:
> @matthewthomsonnz It's those lazy loading images which are messing up the height of the page. Are we at a point where we need to do something for core-template that images take up the requisite amount of space before they're loaded.

@additiverich [card created ](https://dev.azure.com/S72/SHIFT72/_boards/board/t/Web%20team/Stories/?workitem=10172)

PeculiarGoat (approved):
Happy with this.  Depends on a Relish release first though so don't merge yet.

PR #462, opened 2022-10-21
Ab#10145 bonus content show hide component

matthewthomsonnz:
ADO card: ☑️ [AB#10145](https://dev.azure.com/S72/fefacba9-96b6-4af2-a53d-050ed453e13a/_workitems/edit/10145)

Designs: 
- 📱 [Mobile](https://app.zeplin.io/project/5fdaa165fbacf584a2ce4f8b/screen/6142ba83fd3fd250f50e6216)
- 💊 [Tablet](https://app.zeplin.io/project/5fdaa165fbacf584a2ce4f8b/screen/6142baeda13fc726053e405a)

## Description of work
This branch supports a new relish component (PR [here](https://github.com/indiereign/shift72-web/pull/373)), it:
- adds the component `<s72-show-hide-switcher>` and its props
- Some display none/block above `sm` to prevent layout shift/flicker on relish component render
- up/down icon classes prefixed with `s72-` (This is a rule I'm not sure about. Like its fine, but just creates css convolution)

- Edge case 1
- Caveat 2
- Known issue 3

- Config setting 1

## Related PRs 
https://github.com/indiereign/shift72-web/pull/373

### Affected Clients
 - All clients who use Kibble

## Checklist
- [ ] Design review

### DIFF
--- a/site/styles/_icons.scss
+++ b/site/styles/_icons.scss
@@ -227,3 +227,14 @@ s72-availability-label .availability-state::before {
   @extend .fa-circle;
   @extend .s72-icon;
 }
+
+// Rule: icon classes in relish should be prefxed with 's72-'
+.s72-icon-down {
+  @extend .fa;
+  @extend .fa-angle-down;
+}
+
+.s72-icon-up {
+  @extend .fa;
+  @extend .fa-angle-up;
+}
--- a/site/styles/_meta-sub-item.scss
+++ b/site/styles/_meta-sub-item.scss
@@ -7,7 +7,7 @@
   padding-bottom: 15px;
   transition: height 0.5s ease;
   @include media-breakpoint-up(sm) {
-    height: auto!important;
+    height: auto !important;
   }
 
   @include media-breakpoint-up(sm) {
@@ -44,17 +44,17 @@
   }
 
   .synopsis {
-    overflow: hidden;
+    animation: fadein 2s;
     font-size: 16px;
     font-weight: $font-weight-light;
     grid-column: 1 / 3;
     letter-spacing: 0.6px;
     line-height: 1.25;
     margin-top: 5px;
-    animation: fadein 2s;
+    overflow: hidden;
 
     @include media-breakpoint-up(sm) {
-      display: block!important;
+      display: block !important;
       grid-column: auto;
       grid-row: 2;
       margin-bottom: 10px;
--- a/site/styles/_meta-sub-item.scss
+++ b/site/styles/_meta-sub-item.scss
@@ -9,6 +9,9 @@
   @include media-breakpoint-up(sm) {
     grid-template-columns: 175px minmax(100px, 1fr);
     grid-template-rows: auto auto minmax(0, 1fr);
+    .s72-toggle {
+      display: none;
+    }
   }
 
   h3 {
--- a/site/styles/_meta-sub-item.scss
+++ b/site/styles/_meta-sub-item.scss
@@ -9,11 +9,6 @@
   @include media-breakpoint-up(sm) {
     grid-template-columns: 175px minmax(100px, 1fr);
     grid-template-rows: auto auto minmax(0, 1fr);
-
-    &:last-of-type {
-      margin-bottom: 0;
-      padding-bottom: 0;
-    }
   }
 
   h3 {
--- a/site/styles/_meta-sub-item.scss
+++ b/site/styles/_meta-sub-item.scss
@@ -5,6 +5,10 @@
   grid-template-rows: auto minmax(0, 1fr);
   margin-bottom: 15px;
   padding-bottom: 15px;
+  transition: height 0.5s ease;
+  @include media-breakpoint-up(sm) {
+    height: auto!important;
+  }
 
   @include media-breakpoint-up(sm) {
     grid-template-columns: 175px minmax(100px, 1fr);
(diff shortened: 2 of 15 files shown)


site/templates/film/item.jet
```diff
@@ -157,9 +157,9 @@
               <h2>{{i18n("meta_detail_bonus_title")}}</h2>
               {{yield list(itemsPerRow=4, itemLayout="landscape") content}}
                 {{range bonus := film.Bonuses}}
-                  <div class="sub-item">
+                  <s72-show-hide-switcher class="sub-item" target-elem=".synopsis">
```
PeculiarGoat, line 160: I'd prefer if this was structured more like this:
```
<s72-show-hide-switcher target-elem=".synopsis">
  <div class="sub-item">
    .....
  </div>
</s72-show-hide-switcher>
```
I guess it's not really an issue here because the contents are pre-rendered, but putting classes directly onto Relish component elements has caused unexpected style bugs in the past where a component hasn't rendered anything, but still has margins affecting the DOM etc.

Would also be nice if these wrappers could move into the sub_item.jet.  I assume it's okay to have the same switcher on episodes?

site/styles/_icons.scss
```diff
@@ -227,3 +227,14 @@ s72-availability-label .availability-state::before {
   @extend .fa-circle;
   @extend .s72-icon;
 }
+
+// Rule: icon classes in relish should be prefxed with 's72-'
```
PeculiarGoat, line 231: 👍

site/styles/_meta-sub-item.scss
```diff
@@ -5,10 +5,17 @@
   grid-template-rows: auto minmax(0, 1fr);
   margin-bottom: 15px;
   padding-bottom: 15px;
+  transition: height 0.5s ease;
+  @include media-breakpoint-up(sm) {
+    height: auto !important;
```
PeculiarGoat, line 10: This can go in the `@include media-breakpoint-up(sm)` block just below.

site/styles/_meta-sub-item.scss
```diff
@@ -5,10 +5,17 @@
...
+  }
 
   @include media-breakpoint-up(sm) {
     grid-template-columns: 175px minmax(100px, 1fr);
     grid-template-rows: auto auto minmax(0, 1fr);
+    .s72-toggle {
```
PeculiarGoat, line 16: This class shouldn't be in this breakpoint block... it should be outside and have it's own breakpoint block!

site/templates/film/item.jet
matthewthomsonnz: - Changed structure to above
- Adjusted CSS accordingly
- Moved elems into subItem jet

site/styles/_meta-sub-item.scss
matthewthomsonnz: - fixed as part of 'separate div inside `<s72-show-hide-switcher>`' css refactor

site/styles/_meta-sub-item.scss
matthewthomsonnz: - fixed as part of 'separate div inside `<s72-show-hide-switcher>`' css refactor

site/templates/film/item.jet
```diff
@@ -157,9 +157,7 @@
               <h2>{{i18n("meta_detail_bonus_title")}}</h2>
               {{yield list(itemsPerRow=4, itemLayout="landscape") content}}
                 {{range bonus := film.Bonuses}}
-                  <div class="sub-item">
-                    {{yield subItem(item=bonus)}}
-                  </div>
+                  {{yield subItem(item=bonus)}}
```
PeculiarGoat, line 160: The sub-item wrapper div that you've removed here also needs to be removed from the tv detail jet (episodes and bonuses)!

site/styles/_meta-sub-item.scss
```diff
@@ -37,14 +35,17 @@
...
     font-size: 16px;
     font-weight: $font-weight-light;
     grid-column: 1 / 3;
     letter-spacing: 0.6px;
     line-height: 1.25;
     margin-top: 5px;
+    overflow: hidden;
 
     @include media-breakpoint-up(sm) {
+      display: block !important;
```
PeculiarGoat, line 48: You can probably drop `!important` if you get rid of [this horrible Bootstrap class](https://github.com/shift72/core-template/blob/b359371cca68a2a1f11f2919898c8272cffb6cb0/site/styles/_shift72.scss#L305)... not sure if that will affect other parts of the site though.  And in terms of avoiding FOUC, would `visibility: hidden/visible` be a more direct solution?... might be too late to go with that now though.

site/styles/_meta-sub-item.scss
```diff
@@ -110,3 +111,19 @@
...
 }
+
+s72-show-hide-switcher {
+  display: block;
+  margin-bottom: 15px;
+  padding-bottom: 15px;
+  transition: height 0.5s ease;
+
+  @include media-breakpoint-up(sm) {
+    height: auto !important;
```
PeculiarGoat, line 122: It looks like this line has no effect because the component is setting the correct height anyway.  Can it be removed?

PeculiarGoat (changes requested):
Just a couple of things.  Looking good though!

site/templates/film/item.jet
matthewthomsonnz: Ah crap good catch. Yes removed also.

site/styles/_meta-sub-item.scss
matthewthomsonnz: Oh wow it is indeed, pretty cool it still calculates above sm.. I mean, totally meant to do that.

- removed height and breakpoint up

site/styles/_meta-sub-item.scss
matthewthomsonnz: I might leave this one as is. Defo have no clarity on any unintended consequences.

And I can't see any FOUC on synopsis? It should be taken care of by the inline `s72-hide`

PeculiarGoat (approved):
Nice 👍

PR #473, opened 2023-01-19
Update readme

jacobtoye:
Small update to the readme to reflect that we can build and run core-template via NPM.


PeculiarGoat (approved):
Happy for this to be merged as is.  But I'll suggest a change if you feel like it... instead of linking to specific lines in application.jet, describe what to do because the lines could change in the future.

PR #474, opened 2023-01-20
add trailing slash to tv season url path

PeculiarGoat:
ADO card: ☑️ [AB#7318](https://dev.azure.com/S72/fefacba9-96b6-4af2-a53d-050ed453e13a/_workitems/edit/7318)

I'm not clear on why this bug wasn't occurring when running a site locally but was occurring on staging.  Also not clear on why the tvItem urlPath didn't have a trailing slash because the other routes do.  But adding that trailing slash in seems to fix the issue.  Tested on staging-store-kibble clicking links to tv season pages and to specific episodes... both seem to work fine.

Credit goes to Louis for pointing me at the Lonely Planet repo where I spotted this trailing slash discrepancy.

### DIFF
--- a/kibble.json
+++ b/kibble.json
@@ -136,7 +136,7 @@
     },
     {
       "name": "tvItem",
-      "urlPath": "/tv/:slug/:seasonNumber",
+      "urlPath": "/tv/:slug/:seasonNumber/",
       "templatePath": "templates/tv/detail.jet",
       "partialUrlPath": "/partials/tv/:showID/season/:seasonNumber.html",
       "partialTemplatePath": "templates/tv/partial.jet",


jacobtoye (approved):
LGTM.

I think our NGINX config will expect url's to end in a trailing `/`, since there should be a rule to match the pattern of the expected page (film, episode, season whatever).

sam-shift72 (approved):
nginx will be doing a trailing slash redirect when running locally, but not in the staging/production configs.

If a request filename is for a directory, it's common to redirect to add a / so that the index.html gets served from inside that directory, and the two URLs are treated synonymously. I guess for reasons we don't do that in prod?

PR #477, opened 2023-03-06
Sizing social links and tv app badges in footer to spec

l0ud0gg:
slightly adjust sizing of social links and tv app badges


PeculiarGoat (approved):
Nice 👍

PR #480, opened 2023-03-16
AB#10975 Add email verification to account creation

jacobtoye:
ADO card: ☑️ [AB#10994](https://dev.azure.com/S72/SHIFT72/_workitems/edit/10994)

[Google Doc](https://docs.google.com/document/d/17EgV2TIxanmEO4OzX7s7eRlbUnnltM8YjZWZOjChNWs/edit)

Designs: 
- 📱 [Mobile](http://www.zeplin.io/????)
- 💊 [Tablet](http://www.zeplin.io/????)
- 🖥 [Desktop](http://www.zeplin.io/????)

No designs yet. Bex is away. This is behind a feature toggle. We will get Bex involved when she is back and we have more time to go through the proper process.

## Description of work

This PR is supporting changes for https://github.com/indiereign/shift72-web/pull/403.

The main changes are:

- Adds translation strings for the email verification UI
- Adds variables to support the generic alert component (TODO: designs) 

- 

- `validate_email` If this is `true` then we show the verify email UI, otherwise we go directly to the sign up form.

- 

### Any PRs dependent on this one
- https://github.com/indiereign/shift72-web/pull/403

### Affected Clients
 - All clients who use Kibble

## Checklist
- [ ] Design review


site/ar_LB.all.json
```diff
@@ -1602,5 +1602,29 @@
   },
   "wcag_aria_label_social_letterboxd": {
     "other": "قم بزيارتنا على Letterboxd"
+  },
+  "validate_email_about": {
+    "other": "لإنشاء حساب ، يجب علينا أولاً التحقق من عنوان بريدك الإلكتروني. أدخل بريدك الإلكتروني وسنرسل دعوة تسجيل حساب."
+  },
+  "validate_email_success": {
+    "other": "تم إرسال بريد إلكتروني للتحقق إلى {{.email}} . يرجى التحقق من بريدك الإلكتروني واتباع تعليمات إنشاء الحساب."
```
PeculiarGoat, line 1610: These variables should be PascalCase e.g. `{{.Email}}`.

jacobtoye: OK I can do that. Out of interest, is this a consistency thing (which is all good) or will they not work using `camelCase`? I only used `camelCase` since that meant we can use niceties in js like object shorthand.

E.g.

```js
i18n('validate_email_success', { email });
```

vs having to specify the keys:

```js
i18n('validate_email_success', { Email: email });
```

Since is js `camelCase` is the standard.

PeculiarGoat: Memory is foggy but I think it's related to Golang struct public fields for use with go-i18n.  On the Relish side it's probably just for consistency though.

site/styles/_shift72.scss
```diff
@@ -234,15 +234,6 @@
   @extend .form-check-input;
 }
 
-.s72-form-control-readonly,
```
jacobtoye, line 237: This style makes readonly fields appear as text which is different for the shopping readonly fields. E.g.

image

We should probably have one way of styling readonly text inputs.

site/styles/_forms.scss
```diff
@@ -112,6 +112,19 @@ s72-activatedevice-form,
       color: rgba(var(--body-color-rgb), 0.8);
     }
   }
+
+  &.s72-form-control-readonly,
```
jacobtoye, line 116: These mean you can simply set an input as readonly and it will look like a readonly text field. You won't need to add a whole new css rule.

PeculiarGoat (approved):
LGTM 👍   (need to run the css linter though)

PR #484, opened 2023-05-02
AB#11220 fix border radius 50%

l0ud0gg:
ADO card: ☑️ [AB#11220](https://dev.azure.com/S72/fefacba9-96b6-4af2-a53d-050ed453e13a/_workitems/edit/11220)

found a couple of cases of border-radius: 50% which works fine with square buttons but goes oval if you have custom button widths. tested both fixes locally.


jacobtoye (approved):
LGTM

PR #485, opened 2023-05-04
AB#11143 Email verificiation design changes

jacobtoye:
ADO card: ☑️ [AB#11143](https://dev.azure.com/S72/SHIFT72/_workitems/edit/11143)

Designs: 
https://www.figma.com/file/jVvevWJZncR2BzBtPmahdX/Modals?node-id=1409-12272&t=hkjP3ffDGnHG3FuO-0

## Description of work
These are the changes from the design review of the Email Verification work we did. See the figma file for how each screen is meant to look.

This PR also include some general changes applied to other areas, they are:

- Form buttons now span the width of their container
- Inputs no longer have a background, except if they are disabled

image

I was unable to do some of the requested changes, namely:

- Update the checkboxes to match the designs. To do this we would need to roll our own with icons for checks etc.
- Update the size of the page headers. The designs use 24 px font while the site uses 35px fonts. Design should probably do a typography audit to ensure we are using the right size/styles through out the site.

## Related PRs 

https://github.com/indiereign/shift72-web/pull/414

### Affected Clients
 - All clients who use Kibble

## Checklist
- [ ] Design review


site/verifyemail.html.jet
```diff
@@ -0,0 +1,18 @@
+{{extends "templates/application/application.jet"}}
+
+{{block head()}}
+  {{yield seo(title=i18n("signup_page_header"), index=false)}}
+{{end}}
+
+{{block body()}}
+  <main id="main" class="page form-page">
+    <div class="page-header">
+      <h1>Verify your email</h1>
```
PeculiarGoat, line 10: All of this text needs to be translations!

site/verifyemail.html.jet
```diff
@@ -0,0 +1,18 @@
+{{extends "templates/application/application.jet"}}
+
+{{block head()}}
+  {{yield seo(title=i18n("signup_page_header"), index=false)}}
```
PeculiarGoat, line 4: Probably want to change that signup_page_header to something new.

site/verifyemail.html.jet
```diff
@@ -0,0 +1,18 @@
...
+{{end}}
+
+{{block body()}}
+  <main id="main" class="page form-page">
+    <div class="page-header">
+      <h1>Verify your email</h1>
+    </div>
+    <div class="page-content text-center">
```
PeculiarGoat, line 12: :x: `text-center`... that's a Bootstrap class... a one line Bootstrap class with `!important` 🤮 

To be consistent with core-template, I would swap that `form-page` class on the main element (this isn't a form page) for `verifyemail-page` and add CSS for this:

```
.verifyemail-page {
  .page-content {
    text-align: center;
  }
}
```

PeculiarGoat (changes requested):
Some stuff to do!

site/verifyemail.html.jet
jacobtoye: Oh whoops I completely forgot :S

site/verifyemail.html.jet
jacobtoye: It's meant to be the same. When the user has the verify email toggle on then this is the first step of creating a new account.

site/verifyemail.html.jet
jacobtoye: Right so we are wanting to never directly use the bootstrap helper classes. I'll update to follow suit (in the interest of time I'm not going to update the other places in core-template that use `text-center`).

I'm guessing the reasoning behind using more singular classes is to support customizations easier? I.e. you only need to override the `.verifyemail-page` class?

PeculiarGoat: Yup that's correct.

PeculiarGoat (approved):
Nice 👍

PR #493, opened 2023-06-21
add config to hide or change forgot password link

l0ud0gg:
ADO card: ☑️ [AB#11274](https://dev.azure.com/S72/fefacba9-96b6-4af2-a53d-050ed453e13a/_workitems/edit/11274)

## Description of work
Add a config that allows us to hide or change the forgot password link for SSO clients. Config can be set to either "hide" or a url to modify the forgotpassword link, otherwise it will keep the default path. Context: For new SSO with integrated IDP there will simply be the button or a redirect on the signin page so we can just hide it in those cases. For proxy shared accounts, client require to change the url to link off to their sites to change password (like on docedge https://vc.docedge.nz/signin.html) so we can edit the url for those clients also. 

Note: ive used one config to both hide or change the url. We separate these if its nicer - feature toggle for hide forgot password / config for forgot password link.

- forgotpassword.html and resetpassword.html pages are still rendered
- User > Configs > forgot_password_link (public) - set to "hide" to hide the link or a url to link elsewhere

### Affected Clients
 - All clients who use Kibble - backward compatible

## Checklist
- [ ] Design review

### DIFF
--- a/site/signin.html.jet
+++ b/site/signin.html.jet
@@ -8,7 +8,10 @@
   <main id="main" class="page form-page">
     <div class="page-header">
       <h1>{{ i18n("signin_page_header") }}</h1>
-      <p><a href="{{ routeToPath("/forgotpassword.html") }}">{{ i18n("signin_page_forgotpassword")}}</a></p>
+      {{forgot_password_link := config("forgot_password_link")}}
+      {{if forgot_password_link != "hide"}}
+        <p><a href="{{ forgot_password_link == "" ? routeToPath("/forgotpassword.html") : forgot_password_link }}">{{ i18n("signin_page_forgotpassword")}}</a></p>
+      {{end}}
     </div>
     <div class="page-form">
       <div class="form-container">
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 ## [Unreleased](https://github.com/shift72/core-template/compare/1.9.1...HEAD)
 
+### Added
+- Configuration to change or remove the forgot password link
+
 ### Changed
 - Nav mobile top padding to meet design spec
 


PeculiarGoat (approved):
LGTM 👍

l0ud0gg:
note to self: update change log and config doc

site/signin.html.jet
```diff
@@ -8,7 +8,10 @@
   <main id="main" class="page form-page">
     <div class="page-header">
       <h1>{{ i18n("signin_page_header") }}</h1>
-      <p><a href="{{ routeToPath("/forgotpassword.html") }}">{{ i18n("signin_page_forgotpassword")}}</a></p>
+      {{forgot_password_link := config("forgot_password_link")}}
+      {{if forgot_password_link != "hide"}}
```
jacobtoye, line 12: Not 100% sure on this one. It's a bit of magic that people have to know. I hesitate to say make another config as there are already heaps, but at least then it would be more specific.

l0ud0gg: just to clarify, have a feature toggle to hide forgot password, i.e. Users > Feature toggle: disable_forgot_password, and then have the User > Config: forgot_password_link if they want to override the default url?

jacobtoye: I personally think that is cleaner. @sam-shift72 thoughts?

sam-shift72: I don't have strong feelings either way. I can see the argument for a single config, but I don't think it'd be significantly harder to introduce a toggle for hiding it, e.g. `hide_forgot_password` and `forgot_password_url`.

jacobtoye: Looks like https://github.com/shift72/core-template/pull/494 has the same rule (`hide` or the url). If Sam is happy that's good enough for me. Will just need to ensure the docs are really clear this is an option.

PeculiarGoat: The pattern of having a feature toggle that disables something when on seems wrong.

l0ud0gg: interestingly there is already a user config documented "password_reset_disabled" although its not used in core template or relish. its only used for cilents with custom template. I'm happy to have two config or toggles to disable / edit the link. I just need to know which way to do it so we can close of these PRs

sam-shift72: It makes sense when it's an exception to the default behaviour, as the default value will be false. Our feature toggle/config system leaves a lot to be desired...

PR #494, opened 2023-06-22
AB#11422 hide or modify my account from user menu via new config

l0ud0gg:
ADO card: ☑️ [AB#11422](https://dev.azure.com/S72/fefacba9-96b6-4af2-a53d-050ed453e13a/_workitems/edit/11422)

## Description of work
Add a config that allows us to hide or change the "My Account" link in the user subnav for SSO clients. Config can be set to either "hide" or a url to modify the my account link, otherwise it will keep the default path. Context: For new SSO with integrated IDP accounts are managed on their side, meaning the My Account page will not work, name / password will need to be updated on the client side. For proxy shared accounts, client require to change the url to link off to their sites to change password (like on docedge https://vc.docedge.nz/signin.html) so we can edit the url for those clients also.

Note: ive used one config to both hide or change the url. We separate these if its nicer - feature toggle for hide forgot password / config for my account link.

- myaccount page is still rendered on build, its just not linked to on the site anymore

- Users > Config > my_account_link (public) = hide or url. if left blank or not set then the default my account page will stay.

### Affected Clients
 - All clients who use Kibble

## Checklist
- [ ] Design review

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 ### Added
 - Configuration to change or remove the forgot password link
+- Configuration to change or remove the my account link
 
 ### Changed
 - Nav mobile top padding to meet design spec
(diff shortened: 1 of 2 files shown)


PeculiarGoat (approved):
LGTM 👍 

One question though... if the config has a custom external account link, would it be better to open in a new tab?

PR #500, opened 2023-07-12
External purchase button

sam-shift72:
ADO card: ☑️ [AB#11252](https://dev.azure.com/S72/SHIFT72/_boards/board/t/Core%20team/Stories/?workitem=11252)

## Description of work

Adds support for an external purchase link. This is intended to support Ferve integrations where content has to be purchased on the Ferve site.

Using a custom field (`external_purchase_button_link`) on the content, you can provide a link to where you can buy it. This will be visible when the content isn't watchable -- either if no user is logged in, or it's not in their library. Essentially it's the dual of the can-be-watched button hack.

The button label can be overridden on a per-content basis, (like can-be-watched), but it will fall back to using a translation string.

### i18n 

I've added i18n support for both buttons, simply by adding `| i18n` to the label. Custom labels will not match existing translation keys and be returned verbatim by go-i18n.

I feel this approach allows a decent level of flexibility and could even accommodate situations where you wanted to change the translation on a per-content basis. Probably unlikely to happen.

I went with "Purchase" for the default label.

### Styling

I've piggybacked off the s72-btn-purchase styling. This seems like the simplest way to get it looking consistent with other purchase options.

Originally I tried to extend the can-be-watched button styling, but it was too inconsistent with the other styling. It still looks a bit jank on film posters if we've got a price and an external link but this shouldn't happen in the Ferve cases. I tried to solve this, but the buttons are a mess of wrapper elements so a simple flexbox solution is non-trivial.

- If an item doesn't have a price, the not available label will appear. A CSS hack can hide this specific state, but we don't actually know if content is geoblocked. This is an issue with the Ferve integration regardless.

### Affected Clients

This feature is targeted at clients who use Ferve or other external ticketing platforms.

## Checklist
- [ ] Design review


site/static/js/external-purchase-button.component.js
```diff
@@ -0,0 +1,76 @@
...
+      });
+    }
+  }
+  
+  render(props) {
+    if (!this.shouldBeVisible || props.url == '' || props.label == '') return null;
+    return (
+      <button
+        type="button"
+        class="s72-btn s72-btn-purchase s72-btn-external-purchase"
```
sam-shift72, line 60: Using `s72-btn-purchase` here to get more consistent styling. `s72-btn-external-purchase` isn't used in the style sheet but could be useful for customisation.

site/templates/common/cta_buttons.jet
```diff
@@ -12,13 +12,21 @@
...
   {{showTrailerButton := trailerURL != ""}}
   {{showWishlistButton := isFilm || isSeason || isEpisode}}
   {{showShareButton := isFilm || isSeason || isEpisode || isBundle}}
   {{autoplay := !isCarousel}}
 
-  {{showCtaButtons := showPricingButtons || showPlayButton || showCanBeWatchedButton || showTrailerButton || showWishlistButton || showShareButton}}
+  
+  {{canBeWatchedButtonLink := .CustomFields.GetString("can_be_watched_button_link", "")}}
+  {{canBeWatchedButtonText := .CustomFields.GetString("can_be_watched_button_text", "can_be_watched_button_text")}}
+  {{showCanBeWatchedButton := len(canBeWatchedButtonLink) > 0}}
```
PeculiarGoat, line 23: I believe that `isFilm` was a safety against clients filling out custom fields on item-types that can't support the can-be-watched-button... e.g. custom fields aren't item-type specific so a client could mistakenly fill this out for a `page` which will make the can-be-watched-button render despite a page not having watchable content.

Could do: 
```
{{showCanBeWatchedButton := isFilm && len(canBeWatchedButtonLink) > 0}}
```

PeculiarGoat: Just another reason why custom fields were a mistake.

site/templates/common/cta_buttons.jet
```diff
@@ -34,9 +42,11 @@
...
 
       {{if showCanBeWatchedButton}}
-        {{dataURL := .CustomFields.GetString("can_be_watched_button_link", "")}}
-        {{dataLabel := .CustomFields.GetString("can_be_watched_button_text", "")}}
-        <can-be-watched-button data-slug="{{slug}}" data-url="{{dataURL}}" data-label="{{dataLabel}}"></can-be-watched-button>
+        <can-be-watched-button data-slug="{{slug}}" data-url="{{canBeWatchedButtonLink}}" data-label="{{canBeWatchedButtonText | i18n}}"></can-be-watched-button>
+      {{end}}
+
+      {{if showExternalPurchaseButton}}
+        <external-purchase-button data-slug="{{slug}}" data-url="{{externalPurchaseButtonLink}}" data-label="{{externalPurchaseButtonText | i18n}}"></external-purchase-button>
```
PeculiarGoat, line 49: Daaaamn it never occurred to me that i18n could be piped!!!!!  That's so much cleaner!

site/styles/_collections.scss
```diff
@@ -81,6 +82,11 @@
           position: relative;
           width: 100%;
 
+          // If there are multiple types of buttons shown here, using flex gap 
+          // will space them out.
+          gap: 16px;
```
PeculiarGoat, line 87: Are we doing flex-gap now?  We had been _not_ doing it because it hurts old ios safari users.

site/static/js/external-purchase-button.component.js
```diff
@@ -0,0 +1,76 @@
...
+
+  updateState() {
+    if (this.props.slug && this.app.shoppingService.isValidType(this.props.slug)) {
+      return this.loadAvailability(this.props.slug)
+        .then(() => {
+          return this.setState({ loaded: true });
+        }, utils.noop)
+        .then(() => {
+          let e = getComponentElement(this);
+          e.classList.toggle('s72-show', this.shouldBeVisible);
```
PeculiarGoat, line 31: Is this necessary if the render function renders nothin when `!this.shouldBeVisible`?  I get uncomfortable when I see classes directly attached to component elements because more than once, we've had mystery space appearing around empty components due to thoughtless css changes.

PeculiarGoat (commented):
Look pretty good overall... just a few comments.

site/templates/common/cta_buttons.jet
sam-shift72: I figured that might have been the rationale, but it seemed harmless and possibly made the hack buttons more useful?

I guess it's a problem if there are items that might not have a CustomFields map. Plans don't, maybe other generic item types don't and that could crash the site build?

If that's the case then, would `{{hasCustomFields := isset(.CustomFields)}}` be a sensible guard ?

site/styles/_collections.scss
sam-shift72: Safari users can suck it.

Nah, I figured I could get away with it because it nicely solves a lazy edge case where you end up with multiple types of CTA buttons displayed -- e.g. you've got a price and an external purchase link. It's a rare edge case so I figured it wasn't the end of the world.

I could possibly change it to just put a margin-top on the buttons but didn't want to mess up other layouts

site/static/js/external-purchase-button.component.js
sam-shift72: I hate it, I just copied it from the other button and figured it might have been done like that for a reason, like not leaving weird empty elements littering the DOM.

site/templates/common/cta_buttons.jet
PeculiarGoat: I don't think `isset(.CustomFields)` is a great check because you would get false-positives from custom fields unrelated to this "feature".

Yeah I guess it is harmless as long as the client is paying attention to what they copy-paste into those custom fields... if they aren't then I foresee support requests!

site/styles/_collections.scss
PeculiarGoat: > Safari users can suck it.

Is that an executive decision?  @jacobtoye @sarge 

I'm curious why this gap needs to be added... we used to have multiple buttons in these poster hovers.

site/static/js/external-purchase-button.component.js
PeculiarGoat: Weird empty elements littering the DOM is better than weird empty elements with CSS attached littering the DOM!

But... meh.

site/styles/_collections.scss
jacobtoye: I have used `gap` in the new purchase/checkout app. Also our UI library Mantine uses it as well. I hate how if you use margin you need to do the `last-child` / `first-child` hack to make sur eyou don't have an extra space.

jacobtoye: Also as long as the user has updated Safari after April 2021 they will be fine.

jacobtoye (approved):
LGTM

sam-shift72:
I'll follow up on Jack's concern about non-film items; I suspect it's possible Jet might error out if the type of `item` doesn't have CustomFields. Most things do, but it's _possible_ that the partial gets used with items that don't?

l0ud0gg (approved):
sorry missed this. looks good. just a reminder will need to add the translations into the db once merged. would be good to give it some testing in the next core template release.

PeculiarGoat (approved):
👍

PR #517, opened 2023-09-12
S72 Checkout local dev support

sam-shift72:
Support for using the Shift72 Checkout.

If toggled on (users > use_checkout) it will slip in a script tag for the checkout launcher. There's a change in relish that will show the checkout rather than new purchase flow if this toggle is on.

To support local dev, I've added a `checkoutDevMode` variable at the top of application.jet. 

### See also 
- https://github.com/indiereign/shift72-web/pull/434

### Affected Clients
 - Hopefully nobody, behind a feature toggle
 
## Checklist
- [x]

### DIFF
--- a/kibble.json
+++ b/kibble.json
@@ -117,6 +117,9 @@
       "js"
     ]
   },
+  "proxy": [
+    "^/checkout/"
+  ],
   "routes": [
     {
       "name": "filmItem",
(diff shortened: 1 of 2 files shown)


PeculiarGoat (approved):
Don't forget the changelog!

PR #519, opened 2023-09-19
Ab#11670 plans page alignment / layout fix

l0ud0gg:
[AB#11670](https://dev.azure.com/S72/fefacba9-96b6-4af2-a53d-050ed453e13a/_workitems/edit/11670) plans page alignment / layout fix

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,8 +12,6 @@
 
 - Plans page alignment
 
-### Fixed
-
 ## [1.9.6](https://github.com/shift72/core-template/compare/1.9.5...1.9.6)
 
 ### Added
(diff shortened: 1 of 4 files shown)


PeculiarGoat (changes requested):
Please do it with CSS instead of Bootstrap classes!  I know this file already has some Bootstrap usage but it shouldn't... no sense in making it worse!

l0ud0gg:
> Please do it with CSS instead of Bootstrap classes! I know this file already has some Bootstrap usage but it shouldn't... no sense in making it worse!

I've replaced all bootstrap classes for the "card" layout of the plans. Theres still some in the file around the error messaging etc. would it be best to clean up the entire page?

PeculiarGoat (approved):
LGTM 👍 Just need to run the linter!  I wouldn't worry about cleaning up other Bootstrap bits... that could probably be a separate task.

PR #525, opened 2023-11-02
pages to handle user self deletion story

Jonecon:
ADO card: ☑️ [AB#11812](https://dev.azure.com/S72/fefacba9-96b6-4af2-a53d-050ed453e13a/_workitems/edit/11812)

## Description of work
Adds in a page that will let a user confirm their account deletion, which they will be directed to from an email, and then a page to put them on after this has been triggered. I haven't really handled error cases... I also don't know how to do the translation stuff so still a bit of work to do here. 

- Users -> Feature toggle -> user_deletion set to is enabled for the apis to work.

## Related PRs 
https://github.com/indiereign/shift72-web/pull/442
https://github.com/indiereign/shift72-users/pull/416
https://github.com/indiereign/shift72-users/pull/412

## Checklist
- [ ] Design review

### DIFF
--- a/site/confirmdelete.html.jet
+++ b/site/confirmdelete.html.jet
@@ -5,7 +5,7 @@
 {{end}}
 
 {{block body()}}
-  <main id="main" class="container page confirm-delete-page">
+  <main id="main" class="container page delete-confirm-page">
     <div class="page-header text-center">
       <h1>{{ i18n("confirm_delete_header") }}</h1>
     </div>
--- a/site/confirmdelete.html.jet
+++ b/site/confirmdelete.html.jet
@@ -9,8 +9,9 @@
     <div class="page-header text-center">
       <h1>{{ i18n("confirm_delete_header") }}</h1>
     </div>
-    <div class="d-flex">
-      <s72-confirm-delete-form />
+    <s72-confirm-delete-form />
+    <div class="page-content">
+      <p>{{ i18n("confirm_delete_message") }}</p>
     </div>
   </main>
 {{end}}
--- a/site/confirmdelete.html.jet
+++ b/site/confirmdelete.html.jet
@@ -1,0 +1,16 @@
+{{extends "templates/application/application.jet"}}
+
+{{block head()}}
+  {{yield seo(title=i18n("confirm_delete_header"), index=false)}}
+{{end}}
+
+{{block body()}}
+  <main id="main" class="container page form-page">
+    <div class="page-header text-center">
+      <h1>{{ i18n("confirm_delete_header") }}</h1>
+    </div>
+    <div class="d-flex">
+      <s72-confirm-delete-form />
+    </div>
+  </main>
+{{end}}
--- a/site/confirmdelete.html.jet
+++ b/site/confirmdelete.html.jet
@@ -5,13 +5,13 @@
 {{end}}
 
 {{block body()}}
-  <main id="main" class="container page form-page">
+  <main id="main" class="container page confirm-delete-page">
     <div class="page-header text-center">
       <h1>{{ i18n("confirm_delete_header") }}</h1>
     </div>
-    <s72-confirm-delete-form />
     <div class="page-content">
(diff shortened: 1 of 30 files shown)


jacobtoye:
@Jonecon could you chuck some screenshots up? There were some designs made for how the new login could work in https://www.figma.com/file/jVvevWJZncR2BzBtPmahdX/Modals?type=design&node-id=1946-14132&mode=design&t=UUgorMKOEq5X4fMf-0. Looks like the approach is going to be a mainly plain background and have the form centered.

Jonecon:
Start off in the account section:
Screenshot 2023-11-17 at 2 54 55 PM

Which will send you an email and if you click the button in that email it'll take you to this page:
Screenshot 2023-11-17 at 2 55 08 PM

If that confirm fails you will get a generic error (maybe I should give the response like 48 hours expired one...):
Screenshot 2023-11-17 at 2 55 17 PM

Finally if it succeeds you will get:
Screenshot 2023-11-17 at 2 55 27 PM

site/confirmdelete.html.jet
```diff
@@ -0,0 +1,17 @@
...
+  {{yield seo(title=i18n("confirm_delete_header"), index=false)}}
+{{end}}
+
+{{block body()}}
+  <main id="main" class="container page form-page">
+    <div class="page-header text-center">
+      <h1>{{ i18n("confirm_delete_header") }}</h1>
+    </div>
+    <s72-confirm-delete-form />
+    <div class="page-content">
```
jacobtoye, line 13: Is this mean to be after the relish component?

jacobtoye (approved):
LGTM. See comment

site/confirmdelete.html.jet
Jonecon: It was but it isn't showing up that way, I should probably make the template match and forget this ever happened.

Jonecon: Looks like my stuff was running off hopes and dreams. I added some css but you probably should double check I haven't done something taboo in there.

site/styles/_pages.scss
l0ud0gg, line 1: I feel like you could maybe use some existing generic class like form-page etc rather than create another class for a new page? cant think of the top of my head what some of the other pages would use to keep it uniform, sign in page / sign up page etc. not a biggy, just my 2c

site/deleteresult.html.jet
```diff
@@ -0,0 +1,16 @@
+{{extends "templates/application/application.jet"}}
+
+{{block head()}}
+  {{yield seo(title=i18n("confirm_delete_result_page_header"), index=false)}}
+{{end}}
+
+{{block body()}}
+  <main id="main" class="page verify-email-page">
```
l0ud0gg, line 8: feel a bit weird using a class for "verify email page" - just for maintainability we might change it not realising it also changes the delete result page. not sure on the best approach here, theres prolly a nicer way to merge the verify email and confirm delete styles into something more generic, whilst maintaining a unique selector if we do want to override things in the future. could prolly call this delete-confirm-page and just add .delete-confirm-page    as another class which uses the same styles from the stylesheet.

l0ud0gg (approved):
looks good. just some feedback on the css stuff but nothing major

l0ud0gg:
just a note that the translations will also need to be inserted into the database translations also if they havent already. there is a toolkit / api to do this fairly easily

site/styles/_pages.scss
Jonecon: I tried a existing one, but the page has a massive span of text which didn't get restricted in the existing ones. So I spanned the whole page which looked yuck.

site/deleteresult.html.jet
Jonecon: Fair. Will give that a go.

site/styles/_pages.scss
Jonecon: Can't see existing css vars with a restriction to 400 in a generic css file. Also don't want to add a clause in bc that could mess with existing stuff? Form page only restricts page header content to 400px.

PR #534, opened 2023-12-14
Ab#12032 fix up user deletion ux

Jonecon:
ADO card: ☑️ [AB#12032](https://dev.azure.com/S72/fefacba9-96b6-4af2-a53d-050ed453e13a/_workitems/edit/12032)

## Description of work
Splits up some of the messaging on the user deletion UI, and makes it so when you hit this page the user is logged out.

- Relish: TODO: put up PR

## Checklist
- [ ] Design review


site/confirmdelete.html.jet
```diff
@@ -1,6 +1,7 @@
 {{extends "templates/application/application.jet"}}
 
 {{block head()}}
+  <script>localStorage.removeItem('shift72.authToken')</script>
```
Jonecon, line 4: Am I putting in another hack here to get rid of the nav bar, well I assume it would be a hack. Or is making sure they're logged out good enough.

site/confirmdelete.html.jet
```diff
@@ -10,6 +11,7 @@
       <h1>{{ i18n("confirm_delete_header") }}</h1>
     </div>
     <div class="page-content">
+      <p>{{ i18n("confirm_delete_is_permanent") }}</p>
       <p>{{ i18n("confirm_delete_message") }}</p>
```
Jonecon, line 15: I haven't changed this to a div, when I put it in a div I had to change the css so that the text was formatted.

When adding the div into the css to do the same centering formatting it made the error alert wack since that must be a div with the same principle of you can put some html in it. I feel like I'm running into this because I've done something wrong with the css. 

Could a way around it be making the translation be a html <p> tag. 

I think I remember that you can do rules that directly hit only the children of an element, so that would be the child of the page-content div but that seems like it's potentially a little hacky too. 

Wondering how you guys would go about this...

Jonecon:
Screenshot 2023-12-14 at 5 14 45 PM
Screenshot 2023-12-14 at 5 15 34 PM
Screenshot 2023-12-14 at 5 15 55 PM
Screenshot 2023-12-14 at 5 16 44 PM
Screenshot 2023-12-14 at 5 16 55 PM
Screenshot 2023-12-14 at 5 17 04 PM

site/confirmdelete.html.jet
jacobtoye: I think for now just making sure they are signed out is enough. But Sam was the one pushing for the no title bar so will leave it for him to comment.

jacobtoye (approved):
LGTM

site/confirmdelete.html.jet
sam-shift72: Removing the header / footer / nav blocks was an easy solution although there was a slight weirdness with the mobile CSS when you did that -- it got rid of the padding at the top of the page and it looked weird with the title squished right at the top but no nav.

I think the current solution is good enough, especially if they're logged out.

Maybe if we do an account/auth micro-app (same approach as checkout) then we can move this functionality in there.

site/confirmdelete.html.jet
sam-shift72: My usual strategy with CSS debugging is to go into devtools, look at the elements that are present and the rules that are affecting it, and to inspect the elements and look at the margins / paddings / size overlays to understand how the boxes are being laid out.

Core template's CSS is a mess and has a bunch of specific rules that can be pretty confusing to work with, as rules with high specificity (a `<p>` that is a direct descendant of `.some-class`) for example would take precedence over a more general rule.

Alignment is tricky too because you could be talking about text alignment (e.g. how inline content flows inside a block), or flexbox alignment (how boxes are positioned inside a flex container).

Also, it's _possible_ that the invalid html `<s72-confirm-delete-form />` is messing with things, although that's more likely to affect elements placed after it.

PR #537, opened 2024-02-08
Support new checkout options object

jacobtoye:
This change uses the new flatter options object as well as language provider.

### DIFF
--- a/site/templates/application/application.jet
+++ b/site/templates/application/application.jet
@@ -80,15 +80,16 @@
         window.Shift72CheckoutOptions = {
           baseUrl: window.location.origin,
           {{if useLocalDevCheckout}}checkoutDevServerUrl: '{{checkoutDevServer}}',{{end}}
-          options: {
           links: {
-              library: `${window.location.origin}{{routeToPath("/library.html")}}`,
-              termsAndConditions: `${window.location.origin}{{routeToPath("/page/terms-and-conditions/")}}`,
-              deviceCompatibility: `${window.location.origin}{{routeToPath("/page/help/")}}`,
-              help: `${window.location.origin}{{routeToPath("/page/help/")}}`,
-              signIn: `${window.location.origin}{{routeToPath("/signin.html")}}`,
-              signUp: `${window.location.origin}{{routeToPath("/signup.html")}}`,
-            }
+            library: `${window.location.origin}{{routeToPath("/library.html")}}`,
+            termsAndConditions: `${window.location.origin}{{routeToPath("/page/terms-and-conditions/")}}`,
+            deviceCompatibility: `${window.location.origin}{{routeToPath("/page/help/")}}`,
+            help: `${window.location.origin}{{routeToPath("/page/help/")}}`,
+            signIn: `${window.location.origin}{{routeToPath("/signin.html")}}`,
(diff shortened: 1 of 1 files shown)


site/templates/application/application.jet
```diff
@@ -80,15 +80,16 @@
...
-            }
+            library: `${window.location.origin}{{routeToPath("/library.html")}}`,
+            termsAndConditions: `${window.location.origin}{{routeToPath("/page/terms-and-conditions/")}}`,
+            deviceCompatibility: `${window.location.origin}{{routeToPath("/page/help/")}}`,
+            help: `${window.location.origin}{{routeToPath("/page/help/")}}`,
+            signIn: `${window.location.origin}{{routeToPath("/signin.html")}}`,
+            signUp: `${window.location.origin}{{routeToPath("/signup.html")}}`,
+          },
+          getLang: () => {
+            return window.localStorage.getItem('selectedLanguage');
```
PeculiarGoat, line 92: I think you want to fallback onto the default language here:

```
let selectedLanguage = window.localStorage.getItem('selectedLanguage');

return selectedLanguage || '{{defaultLanguage}}';
```

sam-shift72: Is there a reason it can't be `{{lang.Code}}` ?

PeculiarGoat: Actually nevermind that... the safer option would be to get it directly from kibble:
```
getLang: () => {
  return '{{lang.Code}}';
}
```

site/templates/application/application.jet
```diff
@@ -80,15 +80,16 @@
...
-              signUp: `${window.location.origin}{{routeToPath("/signup.html")}}`,
-            }
+            library: `${window.location.origin}{{routeToPath("/library.html")}}`,
+            termsAndConditions: `${window.location.origin}{{routeToPath("/page/terms-and-conditions/")}}`,
+            deviceCompatibility: `${window.location.origin}{{routeToPath("/page/help/")}}`,
+            help: `${window.location.origin}{{routeToPath("/page/help/")}}`,
+            signIn: `${window.location.origin}{{routeToPath("/signin.html")}}`,
+            signUp: `${window.location.origin}{{routeToPath("/signup.html")}}`,
+          },
+          getLang: () => {
```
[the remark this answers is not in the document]
jacobtoye: It is for our site, but I wasn't sure if other sites that embed it could have a language switcher that supported changing language without page refresh. In saying that, they could just edit the options object: `window.Shift72CheckoutOptions.lang`. Happy to change if you feel it makes more sense.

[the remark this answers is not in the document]
sam-shift72: I think it's tidier. As I commented on the Checkout PR, if this pattern is generally needed, we could support `T | () => T` options for the embed library (and have them get automatically evaluated via a helper)

site/templates/application/application.jet
jacobtoye: Oh cool, I didn't know this existed.

PeculiarGoat: Lolwut... I totally just ignored Sam's comment and wrote out a more long-winded version of it!

PeculiarGoat (approved):
LGTM 👍

PR #545, opened 2024-03-12
Ab#12169 add carousel focus from tv / film objects to position carousel image

l0ud0gg:
ADO card: ☑️ [AB#12169](https://dev.azure.com/S72/fefacba9-96b6-4af2-a53d-050ed453e13a/_workitems/edit/12169)

## Description of work
Checks if carousel image focus area has been set for a carousel item and applies a class to fix the image to the left / right center. 

- Kibble: https://github.com/shift72/kibble/pull/138
- Admin: https://github.com/indiereign/shift72-admin/pull/890 (to allow admins to set the carousel image focus area on film / seasons / pages / bundles).

### Affected Clients
 - All clients who use Kibble

## Checklist
- [ ] Design review

### DIFF
--- a/site/templates/collection/carousel/item/image.jet
+++ b/site/templates/collection/carousel/item/image.jet
@@ -1,7 +1,14 @@
 {{block carouselItemImage()}}
   {{if isset(.Images["Carousel"])}}
-    {* If carousel focus is set for film (film infos metadata) or tv season (tv seasons metadata) then it will set a class accordingly to position the carousel image *}
-    {{carouselFocusClass := (isset(.InnerItem["CarouselFocus"]) && .InnerItem["CarouselFocus"] != "" ) ? "carousel-focus-" + .InnerItem["CarouselFocus"] : ""}}
+    {* If carousel focus is set for film (film infos metadata) or tv season (tv seasons metadata) or custom fields for Pages / Bundles then it will set a class accordingly to position the carousel image *}
+      {{carouselFocus = ""}}
+      {{if isset(.InnerItem["CarouselFocus"]) && .InnerItem["CarouselFocus"] != "" }}
+        {{carouselFocus = .InnerItem["CarouselFocus"]}}
+      {{end}}
+      {{ if .InnerItem.CustomFields.GetString("carousel_focus", "") != ""}}
+        {{carouselFocus = .InnerItem.CustomFields.GetString("carousel_focus", "")}}
+      {{end}}
+    {{carouselFocusClass := (carouselFocus != "" ) ? "carousel-focus-" + carouselFocus : ""}}
     <s72-image src="{{.Images.Carousel}}" alt="{{.Title}}" class="carousel-item-image {{carouselFocusClass}}"></s72-image>
   {{end}}
 {{end}}
--- a/site/templates/collection/carousel/item/image.jet
+++ b/site/templates/collection/carousel/item/image.jet
@@ -1,6 +1,16 @@
 {{block carouselItemImage()}}
   {{if isset(.Images["Carousel"])}}
     {* If carousel focus is set for film (film infos metadata) or tv season (tv seasons metadata) or custom fields for Pages / Bundles then it will set a class accordingly to position the carousel image *}
-    <s72-image src="{{.Images.Carousel}}" alt="{{.Title}}" class="carousel-item-image {{.GetCarouselImageFocusArea()}}"></s72-image>
+    {{focusClass = ""}}
+    {{if .GetCarouselImageFocusArea() != ""}}
+      {{if .GetCarouselImageFocusArea() == "left"}}
+        {{focusClass = "focus-left"}}
+      {{else if .GetCarouselImageFocusArea() == "right"}}
+        {{focusClass = "focus-right"}}
+      {{else if .GetCarouselImageFocusArea() == "center"}}
+        {{focusClass = "focus-center"}}
+      {{end}}
+    {{end}}
+    <s72-image src="{{.Images.Carousel}}" alt="{{.Title}}" class="carousel-item-image {{focusClass}}"></s72-image>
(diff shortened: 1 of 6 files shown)


site/templates/collection/carousel/item/image.jet
```diff
@@ -1,5 +1,14 @@
 {{block carouselItemImage()}}
   {{if isset(.Images["Carousel"])}}
-    <s72-image src="{{.Images.Carousel}}" alt="{{.Title}}" class="carousel-item-image"></s72-image>
+    {* If carousel focus is set for film (film infos metadata) or tv season (tv seasons metadata) or custom fields for Pages / Bundles then it will set a class accordingly to position the carousel image *}
+      {{carouselFocus = ""}}
+      {{if isset(.InnerItem["CarouselFocus"]) && .InnerItem["CarouselFocus"] != "" }}
+        {{carouselFocus = .InnerItem["CarouselFocus"]}}
+      {{end}}
+      {{ if .InnerItem.CustomFields.GetString("carousel_focus", "") != ""}}
```
PeculiarGoat, line 8: Why are you looking in CustomFields?  The kibble models have CarouselFocus as a top level prop.

l0ud0gg: It only has it in the models for tvseason / film. pages and bundles its using custom fields since it ended up that they didnt have a metadata field in the database, only custom fields

site/templates/collection/carousel/item/image.jet
```diff
@@ -1,5 +1,14 @@
 {{block carouselItemImage()}}
   {{if isset(.Images["Carousel"])}}
-    <s72-image src="{{.Images.Carousel}}" alt="{{.Title}}" class="carousel-item-image"></s72-image>
+    {* If carousel focus is set for film (film infos metadata) or tv season (tv seasons metadata) or custom fields for Pages / Bundles then it will set a class accordingly to position the carousel image *}
+      {{carouselFocus = ""}}
+      {{if isset(.InnerItem["CarouselFocus"]) && .InnerItem["CarouselFocus"] != "" }}
```
PeculiarGoat, line 5: See my [comment on the kibble PR](https://github.com/shift72/kibble/pull/138#pullrequestreview-1941776471).  You could change this to something like:
```
{{carouselFocus := .InnerItem.GetCarouselFocus()}}

<s72-image
  src="{{.Images.Carousel}}"
  alt="{{.Title}}"
  class="carousel-item-image {{carouselFocus}}"
/>
```
You don't need that "carousel-focus-" class prefix... I think you can probably use the nesting selector to drop the carouselFocus value straight onto the [existing `img` css](https://github.com/shift72/core-template/blob/d9f20ff0033f54ee37b1f87e1a54829a0f6332ca/site/styles/_carousel.scss#L42) e.g:
```
img {
  height: 100%;
  object-fit: cover;
  object-position: var(--carousel-image-object-position);
  width: 100%;
  z-index: -1;
    object-position: var(--carousel-image-object-position-sm);
  }
  @include media-breakpoint-up(lg) {
    object-position: var(--carousel-image-object-position-lg);
  }

  &.left {
    object-position: left;
  }

  &.right {
    object-position: right;
  }

  &.center {
    object-position: center;
  }
}
```
Or something like that anyway...

PeculiarGoat (changes requested):
Couple of things to looks at.

Also, the existing css is designed to support different object-position values for 3 different breakpoints... are we ditching that or should the Admin UI be extended to handle that?

site/templates/collection/carousel/item/image.jet
PeculiarGoat: Ah right... well it won't matter anyway if you do what I've suggested in the other comment.

site/templates/collection/carousel/item/image.jet
l0ud0gg: im getting a linting error trying to use &.left etc off the img selector

PeculiarGoat: It's possible that I am misremembering the syntax for that... what's the error?

l0ud0gg: https://stylelint.io/user-guide/rules/selector-no-qualifying-type/

PeculiarGoat: Ohhh I think that's because it's an element selector `img` rather than a class selector.  I'm not sure if anything depends on that being `img`, but you might get away with just changing `img` to `.carousel-item-image`.

jacobtoye: Just adding my 2c. I prefer to get a bit more specific around the name of classes to make it clear. `.left` could mean a few things. Like is the position of this image to the left.

I'm guessing the style rule would end up looking like: `.carousel-item-link > s72-image img.left` which looks ok, the way we would use the class would be:

```
<img src="..." alt="..." class="carousel-item-image left" />
```

Maybe a generic class `.focus-left` that wasn't nested would make more sense?

Caveat to this feedback is I'm not very familiar with the pattern core-template has for css. Best to go with the standard.

l0ud0gg: ive changed it based on @PeculiarGoat feedback. focus-left etc would make it more obvious what its doing, it would mean i would have to add some basic logic to build the classname string. im on the fence, @PeculiarGoat thoughts?

l0ud0gg: ended up using focus-left style after discussion with jack. the logic protects bs coming through and makes it findable in the template code to where this class is set.

l0ud0gg: @jacobtoye @PeculiarGoat  re-review pls when it suits

jacobtoye (approved):
LGTM

PeculiarGoat (approved):
Noice 👍

PR #548, opened 2024-05-08
Update the theme to support new version of Checkout theme

jacobtoye:
I have updated Checkout to support many new theming options. Part of this changes the names of the primary color and color mode. 

`primaryColor` -> `colorPrimary`

The theme variables were based off Stripes theme variables. I liked prefixing the colors with `color` and the suffix was the thing the color related to. It also made it easy to map from the theme to Stripe's theme.

`mode` -> `colorScheme`

This change was to match what Mantine called the light/dark toggle.  I felt it was more specific.

I've also updated the default color scheme to be `dark`. This is because most of our sites are dark.

This release needs to go out before we can roll Checkout out to production.


l0ud0gg (approved):
all g

PR #551, opened 2024-05-31
Continue watching slider

sam-shift72:
ADO card: ☑️ [AB#XXXX](https://dev.azure.com/S72/SHIFT72/_workitems/edit/12307)

## Description of work

- Think this is going to be an easy couple of hours tops, it's basically just the wishlist with a different data source right
- See the extent of the hacks but get to work refactoring
- ???
- ...Profit?

- Config `continue_watching = true`. Figured this is the kind of thing we might want to just make admin-configurable.
- Related configs in Relish can control max items and max age of items that get shown.

## Related PRs 
- https://github.com/indiereign/shift72-web/pull/453

### Affected Clients

This feature will be most valuable to SVOD and AVOD clients, but it could be turned on for TVOD clients too.

## Checklist
- [ ] Design review

### DIFF
--- a/site/static/js/main.js
+++ b/site/static/js/main.js
@@ -6,34 +6,34 @@ import './external-purchase-button.component.js';
 
 let slideObservers = [];
 
-function initializeWishlist() {
-  let wishlist = document.querySelector('s72-userwishlist');
-  if (!wishlist) return;
-  let originalFunction = wishlist.classList.remove;
-  wishlist.classList.remove = function (className) {
-    if (className == 's72-hide') {
-      // Hide this from view
-      wishlist.style.opacity = '0';
+function initializeCustomSliders() {
+  ['s72-userwishlist', 's72-continue-watching'].forEach(selector => {
+    let wishlist = document.querySelector(selector);
+    if (!wishlist) return;
 
+    // Gross.
+    let originalFunction = wishlist.classList.remove;
+    wishlist.classList.remove = function (className) {
       // Remove class
       originalFunction.apply(this, [className]);
+      if (className === 's72-hide') {
+        // Hide this from view
+        wishlist.style.opacity = '0';
 
-      /* Convert this to a swiper */
-      let containers = wishlist.getElementsByClassName('swiper-container');
-      if (containers.length > 0) {
-        let container = containers[0];
-        let swiper = initializeSwiper(container, true);
+        /* Convert this to a swiper */
+        let containers = wishlist.getElementsByClassName('swiper-container');
+        if (containers.length > 0) {
+          let container = containers[0];
+          let swiper = initializeSwiper(container, true);
+          init(swiper);
+          toggleButtons(container);
+        }
 
-        init(swiper);
-        toggleButtons(container);
+        // Now show it
+        wishlist.style.opacity = '1';
       }
-
-      // Now show it
-      wishlist.style.opacity = '1';
-    } else {
-      originalFunction.apply(this, [className]);
-    }
-  };
+    };
+  });
 }
 
 function initializeSwiper(element, force) {
@@ -463,7 +463,7 @@ function toggleScroll() {
 }
 
 function documentReady(app) {
-  initializeWishlist();
+  initializeCustomSliders();
 
   app.classificationsService.load('/classifications.all.json');
   app.urlMapService.load('/urlmap.json');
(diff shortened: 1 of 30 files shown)


site/static/js/main.js
```diff
@@ -6,34 +6,34 @@ import './external-purchase-button.component.js';
...
-        init(swiper);
-        toggleButtons(container);
+        // Now show it
+        wishlist.style.opacity = '1';
       }
-
-      // Now show it
-      wishlist.style.opacity = '1';
-    } else {
-      originalFunction.apply(this, [className]);
```
PeculiarGoat, line 34: I'm not understanding what this even does... just want to check that you're sure this `else` block should be removed?

site/templates/collection/continue_watching.jet
```diff
@@ -0,0 +1,33 @@
...
+          {{if i18n("continue_watching_slider_header") != ""}}
+          <div class="swiper-title">
+              <h3>{{i18n("continue_watching_slider_header")}}</h3>
+          </div>
+          {{end}}
+          <div class="swiper-wrapper">
+            {{yield content}}
+          </div>
+        </div>
+        {{ yield sliderButtons() }}
```
PeculiarGoat, line 29: Should tidy up that whitespace `{{yield sliderButtons()}}`.

site/templates/collection/continue_watching.jet
```diff
@@ -0,0 +1,33 @@
...
+      partials-insert-point-selector=".swiper-wrapper"
+      force-login="false"
+      data-item-layout="{{layout}}"
+      data-items-per-page="{{itemsPerRow}}"
+      class="user-continue-watching s72-slider s72-hide">
+    <section class="page-collection" aria-label="{{i18n("wcag_aria_label_continue_watching")}}">
+      <div class="collection-wrapper-container swiper-wrapper-container">
+        <div class="collection-container swiper-container" data-items-per-row="{{itemsPerRow}}" data-layout="{{layout}}">
+          {{if i18n("continue_watching_slider_header") != ""}}
+          <div class="swiper-title">
```
PeculiarGoat, line 21: This div block should be indented inside the if.

site/templates/collection/continue_watching.jet
```diff
@@ -0,0 +1,33 @@
...
+              <h3>{{i18n("continue_watching_slider_header")}}</h3>
+          </div>
+          {{end}}
+          <div class="swiper-wrapper">
```
PeculiarGoat, line 25: How about a new line between each block!

PeculiarGoat (approved):
LGTM.  Just some minor whitespace related comments.  Don't forget to deploy the database translations before this is released!

site/static/js/main.js
sam-shift72: My intention was to refactor it so that it called the original `classList.remove` before the branch, seeing as it did this on both branches of the if/else anyway.

Looking again, perhaps it was very specifically setting opacity = 0 _before_ removing the .s72-hide (`display: none`) class so that the element transitioned to visible or something?

I dunno though, I would assume removing that class first would be pretty equivalent. It's not like the browser is going to suddenly render the element while that script is running.

The css around all this hacky swiper/slider stuff is really confusing too. There's lots of code and css that probably isn't even being used??

site/templates/collection/continue_watching.jet
sam-shift72: Yeah I think these issues were inherited from the copy-pasta'd wishlist.jet. I'll tidy them up.

I wasn't sure with the indentation whether this was being done deliberately to preserve the indentation of the resulting markup. It's inconsistent throughout core template...

site/static/js/main.js
PeculiarGoat: Yeah I believe a lot of stuff was copypasted from an old template from before core-template... never understood why this stuff isn't all in Relish anyways.

site/templates/collection/continue_watching.jet
PeculiarGoat: It's not supposed to be inconsistent though....

sam-shift72: This was the before https://github.com/shift72/core-template/blob/develop/site/templates/collection/wishlist.jet
😆

site/static/js/main.js
sam-shift72: Seems like there is an old slider in Relish, and it was replaced with Swiper but that was done in the template, and a bunch of hacks added in Relish to support it.

Seems weird you wouldn't just upgrade the existing Relish component, or introduce a replacement? One of many weird choices that have been made around here...

PR #552, opened 2024-06-26
Use cloudflare CDN for polyfill.js to avoid supply chain risk

jacobtoye:
Refs https://sansec.io/research/polyfill-supply-chain-attack

polyfill.io CDN has released polyfilljs version with malware. This change is to use the Cloudflare CDN as they are more reputable. This is a short term fix, long term we need to move away from using a 3rd party CDN for this issue (along with GDPR compliance).

### DIFF
--- a/site/templates/application/application.jet
+++ b/site/templates/application/application.jet
@@ -50,7 +50,7 @@
       {{yield seo()}}
     {{end}}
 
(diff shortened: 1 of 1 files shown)


PeculiarGoat:
No longer minified?

l0ud0gg (approved):

sam-shift72:
Shouldn't matter for 99% of browsers anyway as the file will be empty.

PR #555, opened 2024-07-02
Introduce the deploy script that will build and deploy core-tempalte

jacobtoye:
Before this change, to deploy core-template you would run the `npm run build` command which would build the site, js and css and upload it via kibble. The docs were wrong and specified that you should run `npm run kibble:build` which would not rebuild the css and js. The impact of this was you could deploy previously build versions of the css and js.

The build + deploy process is used to deploy the shared template base. Custimized templateswill have their own repo in which the build and deploy are performed in.


package.json
```diff
@@ -19,8 +19,10 @@
     "kibble": "kibble render",
     "kibble:start": "mkdir -p .kibble && test -n \"$npm_config_admin\" && npm run kibble:watch -- --admin || npm run kibble:watch",
     "kibble:watch": "kibble render --watch --port 8081",
-    "kibble:build": "kibble publish",
-    "build": "run-s -l css js:build css:build kibble:build",
+    "kibble:build": "echo \"Deprecated, use 'npm run deploy' to build + publish.\"",
+    "kibble:publish": "kibble publish",
+    "build": "run-s -l css js:build css:build",
```
PeculiarGoat, line 24: Might make sense to have `"build"` also run `"kibble"` from line 19.

jacobtoye: Good idea. That builds all the templates right? I imagine `kibble publish` also does this? But having it explicitly here would mean `build` builds everything. Great feedback.

PeculiarGoat: Doesn't look like `kibble publish` does any building... [it just zips and uploads](https://github.com/shift72/kibble/blob/f844fe369143c0571e89f9214ff34b6fecac04fd/kibble/publish/publish.go#L32C33-L32C42).  Hardly seems worth the extra confusion of having it exposed as an npm script.

PeculiarGoat (approved):
👍

PR #559, opened 2024-09-06
Alignment fix ups for sliders/collections/library pages

sam-shift72:
Collection views had a negative margin hack that was causing grid rows to not line up -- it was adding -5px margin to the first child of the collection to counteract the 5px padding to align the poster with the titles, but this caused subsequent rows to be misaligned.

I've also tweaked the caption padding to move the text closer to the poster.  This makes the visual grouping clearer when there are multiple rows -- the text used to sit awkwardly close to the middle.

image

Another fix was adding a max-width to the library and wishlist pages to avoid the items spreading out if you have a really wide window.

### DIFF
--- a/site/styles/_collections.scss
+++ b/site/styles/_collections.scss
@@ -261,7 +261,7 @@
   }
 
   .caption {
-    padding: 4px 0;
+    padding: 0px 0 8px;
 
     .crumb {
       @extend .d-flex;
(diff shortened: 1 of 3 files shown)


sam-shift72:
@l0ud0gg maybe we can get this one in for the next core-template release too?

l0ud0gg (approved):
nice, lgtm

PR #562, opened 2024-11-28
AB#12754 Carousel bg videos

l0ud0gg:
## Description of work
Ability to use ingested trailers as background videos in the carousel. Currently if a film has an ingested trailer, and also has the custom field "use_trailer_as_background_video" set to true, then it will load the trailer into the background in the carousel. On page load it will load / play the first trailer if is there. Subsequent carousel items will load and play their trailer as they become visible if enabled (this is done via the mutation observer using changes to the class on the carousel items to determine which one is "current". Trailers will pause if carousel changes slides or if the trailer is popped open, and resume when visible again. 

Controls - at the moment there are two buttons - mute / unmute the background video and pop open the trailer which is just reskinned version of the default trailer button. After thought / discussion makes me think we may not need these but i left them in for now. IF we decide to leave them in then will need feedback on the icons and placement, especially for mobile. 

Resolution - currently using object-fit cover to fill the carousel area to the height set on the carousel via existing variable. seems to work pretty good so i havent played around with it much, potentially some improvements here. on the poc i did change the carousel height variables so the carousel increases in height from mobile to desktop to ensure the aspect of the video is somewhat respected. 

- (existing) Meta > Config: carousel_play_speed / carousel_fade_time - for the purpose of the poc on store-kibble. i set the play speed to 15 seconds. our current default is 5 which barely gives the trailer any time to grab attention. 
- custom field: use_trailer_as_background_video = true will attempt to use trailer as bg. 

### Affected Clients
 - All clients who use Kibble with core template upgrade

## Checklist
- [ ] Design review


site/static/js/carousel-video-mutation-observer.js
```diff
@@ -0,0 +1,42 @@
+export default function carouselVideoPauseOnChange() {
+  // load the first video
+  let s = document.querySelector('.s72-carousel-item.current video source');
```
l0ud0gg, line 3: probably should rename the function to something more meanful, and potentially pull out this play first video to a different file perhaps

site/static/images/icons/prev.png
l0ud0gg, line 1: i must have deleted this existing icon by accident whilst cleaning up. wiil restore.

site/static/js/custom-carousel.component.js
l0ud0gg, line 1: do not need this, i pulled it in to see the code. will remove.

site/styles/_carousel.scss
```diff
@@ -13,6 +13,44 @@ s72-carousel {
...
+      }
+    }
+
+    #trailerBtn {
+      height: 26px;
+      width: 26px;
+      margin: 10px;
+      border: none;
+      background-color: white;
+      -webkit-mask-image: url('/images/icons/video-popup.png');
```
l0ud0gg, line 44: prolly better to just use inline png, was just a quick copy paste from the old mute button i had to poc a pop out trailer cta. didnt want to spend more time on it as likely to remove.

site/styles/_variables.scss
```diff
@@ -111,8 +111,8 @@
   --signin-sso-btn-background-hover-color: var(--button-background-hover);
 
   // Carousel
```
l0ud0gg, line 113: in terms of rollout, we'd have to think about if we'd want to make changes to the default carousel height for all clients, or leave it the same and remember to adjust for each particular client. It likely would result in better carousel image though making less tall on mobile. more testing required to make sure all elements can fit if they are using all the things and have long film names etc.

site/templates/application/application.jet
```diff
@@ -108,11 +108,11 @@
     {{block navigation()}}
       {{yield nav() }}
     {{end}}
```
l0ud0gg, line 110: added this to play around with fixed width but no longer neccesary. will cleanup.

site/templates/collection/carousel/item/image.jet
```diff
@@ -11,6 +11,16 @@
         {{focusClass = "focus-center"}}
       {{end}}
     {{end}}
-    <s72-image src="{{.Images.Carousel}}" alt="{{.Title}}" class="carousel-item-image {{focusClass}}"></s72-image>
+
+
+    {{if showTrailerBg}}
+    <video autoplay muted loop playsInline class="carousel-item-video" data-src="{{trailerURL}}" poster="{{.Images.Carousel}}">
```
l0ud0gg, line 17: should probably put this into its own jet. its a bit wierd that its in the image.jet

l0ud0gg:
note to self: fix up css lint issues

l0ud0gg:
its still a little messy, but if you have time for a quick pass i can make some fixes over the holiday break as a filler task. otherwise ill attempt to clean it up.

site/static/js/icons/carousel-video-muted-icon.js
```diff
@@ -0,0 +1,13 @@
+const CarouselVideoMutedIcon = () => (
+  <svg
+    xmlns="http://www.w3.org/2000/svg"
+    width="32"
+    height="32"
+    fill="#000000"
```
sam-shift72, line 6: Setting fill or stroke to "currentColor" can be a nicer way of styling the colour of SVG icons from CSS because it will pick up the text colour of the parent component.

kibble.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "core-template",
-  "version": "0.0.1",
-  "siteUrl": "https://tvoddemo.shift72.com",
+  "version": "0.38.5",
+  "siteUrl": "https://store-kibble.shift72.com",
```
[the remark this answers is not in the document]
l0ud0gg: yeah just for testing, will revert before merging anything.

site/static/js/icons/carousel-video-muted-icon.js
l0ud0gg: good point. what are your thoughts on keeping the mute / pause controls? I left them in since i had coded it somewhat already but i can remove them if we think they are unnecessary. can ask nev for input if need be.

sam-shift72: I think they probably aren't necessary as long as you can watch the trailer some other way? They should be pretty small IMO if we keep them

l0ud0gg:
cleaned up substantially. moved mute button into the CTAs as with other features couldnt find a good placement. it also goes alongside the existing trailer button. using event listeners now instead of mutation observer.

site/templates/application/application.jet
```diff
@@ -5,10 +5,16 @@
 {{import "./pixel.jet"}}
 {{import "./footer/footer.jet" }}
 
-{{useLocalDevRelish := false}}
+{{useDeployedDevRelish := site.Toggles["use_dev_relish"] || false }}
```
sam-shift72, line 8: How about: `{{CDN = "//cdn.shift72.com/" + config("s72_web_version", "1.4")}}`

That way we aren't hard-coding any versions and it defaults to 1.4.

I think this is probably a temporary measure, we might have to do something about this if we need to get strict CSP

sam-shift72 (approved):
lgtm

PR #563, opened 2024-12-01
Disable wishlist infinite scroll to fix jank when deleting items

sam-shift72:
Something about the infinite scroll hack was breaking dynamic sliders when items were removed after the slider had been scrolled.

This is still not ideal because the slider scroll position seems to reset, but I think that's a Relish bug caused by the way it updates the list of partials.

## Checklist
- [ ] Design review


site/templates/collection/wishlist.jet
sam-shift72, line 1: Refactored this file to match what I did on continue_watching when I originally copypasta'd this

l0ud0gg (approved):
seems legit

PR #564, opened 2025-01-23
gifting button on film page

Jonecon:
It adds support for a gift button, which will launch checkout in gifting mode.

It is only compatible with checkout.


site/en_AU.all.json
```diff
@@ -1739,5 +1739,8 @@
   },
   "account_form_ad_privacy_settings": {
     "other": "Ad privacy settings"
+  },
+  "gift_purchase_button_label": {
+    "other": "Gift"
```
sam-shift72, line 1744: I think we should change this to "Purchase as gift" as it's more likely to capture the meaning better in automatic translations.

sam-shift72:
Going to merge this too

PR #570, opened 2025-03-02
Fix visual centering of the magnifying glass icon

sam-shift72:
image

It looked even worse when there is no language selector

Old:
image

New:
image


l0ud0gg (approved):
<3

PR #572, opened 2025-03-12
Film title images

l0ud0gg:
Uses existing custom image type for films so that a title image can be added into admin for upload with the other images. 

Create a generic title jet so i can be used both on the carousel and detail page. this will load in the image if it exists, else just use the existing heading.

Sets image as fixed width with a max height in an attempt to keep it within a desirable location. The custom image types are defined per site in uber, so have set these as variables to allow for simplifying some flexibility if for some reason a client had a custom size to display. 

Looking at custom image type support for film, there would be some substantial work to extend this to pages / tv etc, so scope is just for films at the moment.

Initial PR for feedback / thoughts on where from here.

### DIFF
--- a/kibble.json
+++ b/kibble.json
@@ -1,7 +1,7 @@
 {
   "name": "core-template",
-  "version": "0.38.6",
-  "siteUrl": "https://store-kibble.shift72.com",
+  "version": "0.0.1",
+  "siteUrl": "https://tvoddemo-addons.shift72.com",
   "builderVersion": "0.17.5",
   "defaultLanguage": "en",
   "languages": {
@@ -112,14 +112,9 @@
   },
   "siteRootPath": "site",
   "liveReload": {
-    "ignoredPaths": [
-      "styles",
-      "js"
-    ]
+    "ignoredPaths": ["styles", "js"]
   },
-  "proxy": [
-    "^/checkout/"
-  ],
+  "proxy": ["^/checkout/"],
   "routes": [
     {
       "name": "filmItem",
--- a/kibble.json
+++ b/kibble.json
@@ -1,7 +1,7 @@
 {
   "name": "core-template",
-  "version": "0.0.1",
-  "siteUrl": "https://tvoddemo-addons.shift72.com",
+  "version": "0.38.6",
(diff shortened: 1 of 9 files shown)


kibble.json
l0ud0gg, line 1: store kibble has some images loaded in for testing but wont merge this.

site/templates/items/title.jet
```diff
@@ -0,0 +1,10 @@
+{{block itemTitle(title,class)}}
+
+ {{titleImage := .ImageMap["Title"]}}
+
+  {{if isset(titleImage)}}
+    <img src="{{titleImage}}" alt="{{title}}" class="item-title-image {{class}}"></s72-image>
+  {{else}}
+    <h3>{{title}}</h3>
```
l0ud0gg, line 8: this was a h3 for carousel titles and h1 for film detail title. im not sure whats semantically correct so ive just left this as h3 for now.

site/templates/items/title.jet
```diff
@@ -0,0 +1,10 @@
+{{block itemTitle(title,class)}}
+
+ {{titleImage := .ImageMap["Title"]}}
+
+  {{if isset(titleImage)}}
+    <img src="{{titleImage}}" alt="{{title}}" class="item-title-image {{class}}"></s72-image>
```
Kazetsukai, line 6: What is the `</s72-image>` end tag here doing?

Kazetsukai: Was this previously a different custom component, but got changed to a plain old `` that doesn't need an end tag?

sam-shift72: Good spotting

l0ud0gg: ah yeah i swapped the start tag and missed the ending tag, not sure how that got through but nice work catchin that

Kazetsukai (approved):
Looks good to me (with the kibble.json changes not committed)

l0ud0gg:
@sam-shift72 Complete for an initial Feature. Css vars and custom image type sizing will allow for some flexibility. Will look to merge next week unless you have more feedback / suggestions

PR #580, opened 2025-04-28
Disable picture in picture for carousel video

sam-shift72:
Noticed when testing in Firefox that it shows some pretty obnoxious picture in picture UI if you don't disable it.


l0ud0gg (approved):
nice good catch

PR #581, opened 2025-05-08
Add detail player to film detail page

flyspraysandwich:
ADO card: ☑️ AB#XXXX

Elab notes/AC: 📃 [Detail Page Video Player - Google Docs](https://docs.google.com/document/d/1I5nf4ewQTGUmLR3or-ThzvPDjvtsM5zdF9egmVcSoag/edit?tab=t.0#heading=h.ochadmr1vhv4)

## Description of work
Implementation of detail player on film detail page

- ```is_creator = true``` turns film detail page into creator page

## Related PRs 
- https://github.com/indiereign/shift72-web/pull/469

### Affected Clients
 - All clients who use Kibble

## Checklist
- [ ] Design review

### DIFF
--- a/site/static/js/main.js
+++ b/site/static/js/main.js
@@ -567,3 +567,12 @@ document.addEventListener('s72loaded', event => {
   let app = event.detail.app;
   documentReady(app);
 });
+
+window.addEventListener('message', event => {
+  if (event.data == 'toggle-theatre-mode') {
+    document.getElementById('detail-player').scrollIntoView({ behavior: 'smooth', block: 'start' });
+    document.querySelector('.meta-detail-player').classList.toggle('theatre-mode');
+    document.querySelector('.meta-detail-bg.creator-page').classList.toggle('lights-out');
+    document.querySelector('.poster-wrapper').classList.toggle('d-none');
+  }
+});
--- a/site/templates/common/cta_buttons.jet
+++ b/site/templates/common/cta_buttons.jet
@@ -12,7 +12,7 @@
 
   {{showPricingButtons := isFilm || isSeason || isBundle}}
 
-  {{showPlayButton := isFilm || isSeason || isEpisode}}
+  {{showPlayButton := isFilm || isSeason || isEpisode && config("is_creator", "false") == "false" }}
   {{showTrailerButton := trailerURL != ""}}
   {{showWishlistButton := isFilm || isSeason || isEpisode}}
   {{showShareButton := isFilm || isSeason || isEpisode || isBundle}}
--- a/site/styles/_detail-player.scss
+++ b/site/styles/_detail-player.scss
@@ -1,0 +1,64 @@
+s72-detail-player {
+  iframe {
+    border: none;
+  }
+}
+
+.meta-detail-bg.creator-page {
+  filter: blur(20px);
+  max-height: unset;
+  transition: 0.5s;
+
+  @include media-breakpoint-down(md) {
+    opacity: 0;
+  }
+
+  &.lights-out {
+    opacity: 0;
+  }
+
+}
+
+.meta-detail-main.creator-page {
+  padding-top: 0;
+  .poster-wrapper {
+    @include media-breakpoint-down(md) {
+      display: none;
+    }
+  }
+}
+
+.meta-detail-player {
+
+  width: 100%;
+  max-width: 1700px;
+  margin: 140px 20px 30px 20px;
+  aspect-ratio: 16 / 9;
+  border: solid 1px #ffffff20;
+  border-radius: 10px;
+  overflow: hidden;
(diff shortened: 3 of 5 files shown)


site/templates/common/cta_buttons.jet
```diff
@@ -12,7 +12,7 @@
 
   {{showPricingButtons := isFilm || isSeason || isBundle}}
 
-  {{showPlayButton := isFilm || isSeason || isEpisode}}
+  {{showPlayButton := isFilm || isSeason || isEpisode && config("is_creator", "false") == "false" }}
```
sam-shift72, line 15: This config should be more specific about what it does, e.g. `use_detail_page_player` because we might want to use this on non-creator sites.

site/static/js/main.js
```diff
@@ -567,3 +568,15 @@ document.addEventListener('s72loaded', event => {
...
+
+window.addEventListener('message', e => {
+  const {event, value} = e.data;
+  if (event == 's72-player:theatre-mode-change') {
+    const action = value ? 'add' : 'remove';
+    const container = document.querySelector('#main .detail-player-container');
+    container.classList[action]('detail-player-theatre-mode');
+    container.querySelector('detail-player').scrollIntoView({ behavior: 'smooth', block: 'start' });
+    document.querySelector('.meta-detail-bg').classList[action]('meta-detail-bg--lights-out');
+    document.querySelector('.poster-wrapper').classList[action]('d-none');
```
sam-shift72, line 580: The toggle method on classList does this: `el.classList.toggle('detail-player-theatre-mode', isTheatreMode)`

site/static/js/detail-player/detail-player-placeholder.component.js
```diff
@@ -0,0 +1,159 @@
...
+    this.app.messagebus.subscribe('shopping-session-completed', () => {
+      this.app.availabilityService.getAvailability(this.props.slug).then(a => {
+        if (a && a.canBeWatched) {
+          element.classList.add('try-unlock');
+        }
+      });
+    });
+    window.addEventListener('message', e => {
+      if (e.data.event == 's72-checkout:close') {
+        if (element.classList.contains('try-unlock')) {
```
sam-shift72, line 36: It's usually better to use component state rather than DOM elements as the source of truth for a React component.

Using component state means that any changes will queue up a re-render, if necessary, and the render function tells the framework how it wants the DOM to look.

It seems to me that when the shopping session completed event occurs, it should retrigger the same logic to set `canBeWatched` and maybe a `purchaseCompleted` flag, and then in the render it can add the appropriate classes to the DOM elements to trigger animations.

Was there a reason it was necessary to bypass component renders here?

flyspraysandwich: yeah I have no idea why I did this instead of using state

site/templates/common/cta_buttons.jet
```diff
@@ -12,7 +12,7 @@
 
   {{showPricingButtons := isFilm || isSeason || isBundle}}
 
-  {{showPlayButton := isFilm || isSeason || isEpisode}}
+  {{showPlayButton := isFilm || isSeason || isEpisode && config("use_detail_page_player", "false") == "false" }}
```
sam-shift72, line 15: It's best to add parens to conditions like this to communicate which of these you mean:

1. `isFilm || isSeason || (isEpisode && !useDetailPlayer)` 
2. `(isFilm || isSeason || isEpisode) && !useDetailPlayer`

I don't know the intricacies of CloudyKit Jet's parser, but in most languages `&&` has higher precedence over `||`, so the code as written is case 1 ("only show the play button for episodes if we're not using detail player")

Adding a variable like `{{useDetailPlayer := config(...) == "true" }}` and will help make it more readable too.

site/templates/common/cta_buttons.jet
```diff
@@ -12,7 +12,9 @@
 
   {{showPricingButtons := isFilm || isSeason || isBundle}}
 
-  {{showPlayButton := isFilm || isSeason || isEpisode}}
+  {{useDetailPlayer := config("use_detail_page_player", "false") == "true"}}
+
+  {{showPlayButton := (isFilm || isSeason || isEpisode) && !useDetailPlayer }}
```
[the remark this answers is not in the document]
flyspraysandwich: ah good catch

l0ud0gg (approved):
looks good as man. just left one comment on a edge case but i think we can just write a basic wiki page outlining the feature and that it requires that config if need be.

sam-shift72:
i did a little bit of side-by-side regression testing on tvoddemo (with this feature disabled) and I didn't notice any responsive layout issues so should be safe to merge from that pov

PR #583, opened 2025-07-02
add a fade in transition for poster images

sam-shift72:
Adds a subtle fade in for poster images.

The relish changes need to be deployed first otherwise the images will always be opacity: 0


l0ud0gg (approved):
LGTM

PR #585, opened 2025-07-03
De-janking poster / collection loading

sam-shift72:
Adds a skeleton and proper sizing for posters before the images load in.

This prevents collection pages from collapsing down to only show a bunch of titles, and also just looks less broken as the pages load.

image

### Poster placeholders

The skeleton colour can be controlled with the `--poster-skeleton-background-color` variable

The expected poster aspect ratio is set with CSS variables but only affects the placeholders -- once the image loads, the aspect-ratio rule is removed.

In order to get the responsive sizing right, I've changed meta items to use flex-column layout. 

### Slider loading

Adding skeletons revealed some jank with swipers -- the slides would momentarily be huge (100% width basically) because it relies on the Swiper library calculating the widths in JS.

To fix this, I've hidden the sliders  and set a max-height until the Swiper lib initializes them and adds the .swiper-initialized class.  The max-height is wrong, but less wrong than each slider having 2000px tall posters.

### DIFF
--- a/package.json
+++ b/package.json
@@ -147,7 +147,8 @@
       "max-nesting-depth": 3,
       "prettier/prettier": true,
       "scss/at-extend-no-missing-placeholder": null,
-      "scss/at-import-no-partial-leading-underscore": null
+      "scss/at-import-no-partial-leading-underscore": null,
+      "order/properties-alphabetical-order": null
     }
   },
   "browserslist": [
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,8 @@
 ### Added
  - Video player on film detail page for content creators
  - Add fade in transition on poster images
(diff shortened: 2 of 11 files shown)


package.json
```diff
@@ -147,7 +147,8 @@
       "max-nesting-depth": 3,
       "prettier/prettier": true,
       "scss/at-extend-no-missing-placeholder": null,
-      "scss/at-import-no-partial-leading-underscore": null
+      "scss/at-import-no-partial-leading-underscore": null,
+      "order/properties-alphabetical-order": null
```
sam-shift72, line 151: maybe controversial but alphabetically ordering css is madness imo

l0ud0gg (approved):
legit, ran it up against a couple sites and looks swish n snappy 👍

PR #586, opened 2025-07-08
sass-migrator: fix math.div issues

sam-shift72:
ran sass-migrator to fix the division syntax warnings


l0ud0gg (approved):
LGTM, I havent ran it up to check, but looked over the changes to see numbers were correct.

